PR #12, opened 2024-05-02
Fix the naming of heartbeat_to_dead

zhenthebuilder:
Why
===
Consolidate with what we are using in bun

What changed
============
Rename

Test plan
=========
Unit + babel


blast-hardcheese:
@zhenthebuilder Can this branch be deleted?

PR #14, opened 2024-05-03
0.1.18: Simplify session management

zhenthebuilder:
Why
===
Simplify the session management logic to prevent potential bugs

What changed
===
1. Session create a merged into one function for client
2. Fix some dead locks in the following situations: _send_transport_message could trigger a websocket close callback, which requires `self._ws_lock`, so we shouldn't call _send_transport_message within _ws_lock context, that will trigger a deadlock at the client.

### DIFF
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -112,6 +112,7 @@ class ClientTransport(Transport):
                     if not existing_session
                     else existing_session.session_id
                 )
+                logging.error(f"##### existing session : {existing_session}")
                 rate_limit.consume_budget(client_id)
                 handshake_request, handshake_response = await self._establish_handshake(
                     self._transport_id, self._server_id, session_id, ws
@@ -173,12 +174,16 @@ class ClientTransport(Transport):
             return session_to_replace_ws
 
     async def _get_or_create_session(self) -> ClientSession:
+        logging.error(f"####### _get_or_create_session")
+
         async with self._create_session_lock:
             existing_session = await self._get_existing_session()
             if existing_session:
                 if await existing_session.is_websocket_open():
+                    logging.error(f"####### use existing session")
                     return existing_session
                 else:
+                    logging.error(f"####### retry session connection")
                     session = await self._retry_session_connection(existing_session)
                     # This should never happen, adding here to make mypy happy
                     if not isinstance(session, ClientSession):
@@ -188,6 +193,7 @@ class ClientTransport(Transport):
                         )
                     return session
             else:
+                logging.error(f"####### establish new connection")
                 new_ws, hs_request, hs_response = await self._establish_new_connection()
                 advertised_session_id = hs_response.status.sessionId
                 if not advertised_session_id:
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -57,35 +57,26 @@ class ClientTransport(Transport):
         self._rate_limiter = LeakyBucketRateLimit(
             transport_options.connection_retry_options
         )
-        # We want to make sure there's only one session creation at a time
-        self._create_session_lock = asyncio.Lock()
-        # Only one retry should happen at a time
-        self._retry_ws_lock = asyncio.Lock()
-
-    async def _on_session_closed(self, session: Session) -> None:
-        logging.info(f"Client session {session.advertised_session_id} closed")
-        await self._delete_session(session)
 
     async def close(self) -> None:
         self._rate_limiter.close()
         await self._close_all_sessions()
 
     async def _get_existing_session(self) -> Optional[ClientSession]:
-        async with self._session_lock:
-            if not self._sessions:
-                return None
-            if len(self._sessions) > 1:
-                raise RiverException(
-                    "session_error",
-                    "More than one session found in client, should only be one",
-                )
-            session = list(self._sessions.values())[0]
-            if isinstance(session, ClientSession):
-                return session
-            else:
-                raise RiverException(
-                    "session_error", f"Client session type wrong, got {type(session)}"
-                )
+        if not self._sessions:
+            return None
+        if len(self._sessions) > 1:
+            raise RiverException(
+                "session_error",
+                "More than one session found in client, should only be one",
+            )
+        session = list(self._sessions.values())[0]
+        if isinstance(session, ClientSession):
+            return session
+        else:
+            raise RiverException(
+                "session_error", f"Client session type wrong, got {type(session)}"
+            )
 
     async def _establish_new_connection(
         self,
@@ -112,7 +103,9 @@ class ClientTransport(Transport):
                     if not existing_session
                     else existing_session.session_id
                 )
-                logging.error(f"##### existing session : {existing_session}")
+                logging.error(
+                    f"##### _establish_new_connection: existing session : {existing_session}"
+                )
                 rate_limit.consume_budget(client_id)
                 handshake_request, handshake_response = await self._establish_handshake(
                     self._transport_id, self._server_id, session_id, ws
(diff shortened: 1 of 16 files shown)


replit_river/client_transport.py
```diff
@@ -126,92 +119,60 @@ async def _establish_new_connection(
...
+            handlers={},
+            close_session_callback=self._delete_session,
+            retry_connection_callback=lambda x: self._get_or_create_session(),
+        )
+
+        await self._set_session(new_session, acquire_lock=False)
+        await new_session.start_serve_responses()
+        return new_session
 
     async def _get_or_create_session(self) -> ClientSession:
```
jackyzha0, line 152: nice this is very clean now

replit_river/session.py
```diff
@@ -128,6 +120,7 @@ async def _begin_close_session_countdown(self) -> None:
         self._close_session_after_time_secs = (
             await self._get_current_time() + grace_period_ms / 1000
         )
+        await self.close_websocket(self._ws_wrapper, should_retry=not self._is_server)
```
jackyzha0, line 123: we shouldn't retry when we disconnect due to session timeout

replit_river/session.py
```diff
@@ -164,8 +157,15 @@ async def _handle_messages_from_ws(
             self._ws_wrapper.id,
         )
         try:
-            async for message in self._ws_wrapper.ws:
+            ws_wrapper = self._ws_wrapper
+            async for message in ws_wrapper.ws:
                 try:
+                    logging.error(
```
jackyzha0, line 163: is this debug?

replit_river/session.py
```diff
@@ -267,18 +267,25 @@ async def _heartbeat(
...
+                logging.debug(
+                    "heartbeat right now : {} {} {}".format(
+                        self._heartbeat_misses,
+                        self._transport_options.heartbeats_until_dead,
+                        self._transport_options.heartbeat_ms,
+                    )
+                )
                 if (
                     self._heartbeat_misses
                     >= self._transport_options.heartbeats_until_dead
```
jackyzha0, line 279:
```suggestion
                    > self._transport_options.heartbeats_until_dead
```

after https://github.com/replit/river/commit/8f796736a54bb09939208637b77b9b055cc30ce4

replit_river/transport.py
```diff
@@ -45,13 +45,22 @@ async def _close_all_sessions(self) -> None:
         await asyncio.gather(*tasks)
         logging.info(f"Transport closed {self._transport_id}")
 
-    async def _delete_session(self, session: Session) -> None:
-        async with self._session_lock:
+    async def _delete_session(
+        self, session: Session, acquire_lock: bool = True
+    ) -> None:
+        if acquire_lock:
+            async with self._session_lock:
```
jackyzha0, line 52: why do we need a lock for these two?

jackyzha0 (approved):

replit_river/session.py
zhenthebuilder: this it the begin of session timeout, starting the countdown, we should retry here right?

replit_river/session.py
zhenthebuilder: Yes, removed

replit_river/session.py
zhenthebuilder: Done

replit_river/transport.py
zhenthebuilder: Actually not needed in the context of asyncio

replit_river/session.py
jackyzha0: oh wait you are right yes, and its fine that this potentially races with another reconnect?

replit_river/session.py
```diff
@@ -267,18 +264,25 @@ async def _heartbeat(
                     ACK_BIT,
                 )
                 self._heartbeat_misses += 1
+                logging.debug(
+                    "heartbeat right now : {} {} {}".format(
```
bheilbrun, line 268: minor perf optimization. using `printf` style formatting should be cheaper than `String.format()` since the former won't do any formatting if the log level is INFO level or higher.

replit_river/session.py
```diff
@@ -267,18 +264,25 @@ async def _heartbeat(
...
+                        self._heartbeat_misses,
+                        self._transport_options.heartbeats_until_dead,
+                        self._transport_options.heartbeat_ms,
+                    )
+                )
                 if (
                     self._heartbeat_misses
-                    >= self._transport_options.heartbeats_until_dead
+                    > self._transport_options.heartbeats_until_dead
```
bheilbrun, line 276: during the initial code review i almost commented on this name because it's slightly unclear if this test should be `>` or `>=` and it looks like we tripped over it, heh. `max_failed_heartbeats` or `failed_heartbeat_limit` are probably a bit more common as names. Totally optional, just a thought for future changes.

replit_river/session.py
```diff
@@ -353,13 +357,14 @@ async def send_message(
...
-                    if await self._ws_wrapper.is_open():
-                        # if it is not open it's fine, we already put it the buffer
-                        await self._send_transport_message(
-                            msg,
-                            self._ws_wrapper.ws,
-                            prefix_bytes=self._transport_options.get_prefix_bytes(),
-                        )
+                    if not await self._ws_wrapper.is_open():
+                        # If the websocket is closed, we should not send the message
+                        return
```
bheilbrun, line 362: The Websocket API throws an exception if you `send()` on a closed socket. Is there a reason we want to silently drop the message instead? Just an abstract concern, I don't have enough context to say it's an actual problem.

replit_river/session.py
```diff
@@ -399,9 +404,12 @@ async def close_websocket(
     ) -> None:
         """Mark the websocket as closed, close the websocket, and retry if needed."""
         async with self._ws_lock:
+            # Already closed.
+            if not await ws_wrapper.is_open():
+                return
             await ws_wrapper.close()
         if should_retry and self._retry_connection_callback:
-            await self._retry_connection_callback(self)
+            await self._task_manager.create_task(self._retry_connection_callback(self))
```
bheilbrun, line 412: IIUC, `create_task` doesn't need to be async.

replit_river/session.py
```diff
@@ -491,9 +499,10 @@ async def close(self, is_unexpected_close: bool) -> None:
                 return
             self._state = SessionState.CLOSING
             self._reset_session_close_countdown()
-            await self.close_websocket(self._ws_wrapper, should_retry=False)
+            async with self._ws_lock:
+                await self._ws_wrapper.close()
```
bheilbrun, line 503: I notice in WebsocketWrapper.close() that it sets `ws_state` to `CLOSING` and then `CLOSED`, but no one would ever observe `CLOSING` because there's no `await` call between the two states. We don't wait for the websocket to actually close. I think that's fine, but FYI.

bheilbrun (approved):
LGTM. Non-critical comments.

replit_river/session.py
jackyzha0: we buffer anyways so its ok to not send altogether

jackyzha0:
side note: do we not run tests in ci?

replit_river/session.py
zhenthebuilder: ack

replit_river/session.py
zhenthebuilder: Added a TODO to do that, right now i prefer to keep it consistent with the river protocol to prevent confusion.

replit_river/session.py
zhenthebuilder: very true

zhenthebuilder:
We are running tests, changed the ci naming to build and test

replit_river/session.py
zhenthebuilder: the reconnects are all controlled by the lock in transport, so there will be only one at a time

replit_river/session.py
zhenthebuilder: i think this is fine for now, i don't see us want to observe closed at any degree.

PR #15, opened 2024-05-06
0.1.20: Don't log error if websocket is closed normally before receiving a handshake

zhenthebuilder:
Why
===
There are handshakes come and close immediately, we should not log them as error.

What changed
============
Fix it


lhchavez (approved):
looks cromulent!

jackyzha0 (approved):

PR #20, opened 2024-05-17
Fix unable to pass handshake metadata to Client

masad-frost:
Why
===

In #17 I didn't realize that ClientTransport is constructed internally in Client, so there's currently no way of passing it to ClientTransport!

derp

What changed
============
Add `handshake_metadata` argument to `Client`

Test plan
=========

Same as #17


zhenthebuilder (approved):

masad-frost:
image

PR #23, opened 2024-05-24
Add default value to optional handshake request field

masad-frost:
Why
===

Before https://github.com/replit/river/pull/131, we used to cast `undefined` fields to `null` over the wire (in msgpack). Pydantic throws if there's an `Optional` field without a default value and not specified in the data dict.

Hopefully this Repl explains what's going on https://replit.com/@replitfaris/pydantic-handshake-missing-optional-metadata#main.py

What changed
============

Set default value for `ControlMessageHandshakeRequest['metadata']`.

I think this fix is better than reverting https://github.com/replit/river/pull/131, it's also a fix forward for existing clients. That said we should think about this a little more deeply, because in python-land we have the same problem where `Optional` fields can end up being `null` over the wire.

Test plan
=========

See repl above.

### DIFF
--- a/replit_river/rpc.py
+++ b/replit_river/rpc.py
@@ -51,7 +51,7 @@ class ControlMessageHandshakeRequest(BaseModel):
     type: Literal["HANDSHAKE_REQ"] = "HANDSHAKE_REQ"
     protocolVersion: str
     sessionId: str
-    metadata: Optional[Any]
+    metadata: Optional[Any] = None
 
 
 class HandShakeStatus(BaseModel):


masad-frost:
> I think this fix is better than reverting https://github.com/replit/river/pull/131, it's also a fix forward for existing clients. That said we should think about this a little more deeply, because in python-land we have the same problem where `Optional` fields can end up being `null` over the wire.

So turns out we have `exclude_none=True` on our dump https://github.com/replit/river-python/blob/beaf4a725e07b8ca3a28d055adced62c30e55c54/replit_river/messages.py#L47. While we're in a better spot than i expected, there's one edge case where this is problematic, when a type (in json schema) is a union of `string` (or any other type) and `null`, we'll omit the field in the python implementation :cry:

masad-frost:
Maybe the answer is to ban null from the schemas

jackyzha0 (approved):

PR #28, opened 2024-06-13
Make `Optional` types truly Optional

lhchavez:
Why
===

The codegen currently correctly emits optional data types as `Optional[T]`. That's good, but what was missing was that these should also have a default value of `None`
[[docs](https://docs.pydantic.dev/latest/concepts/fields/)].

What changed
============

This change now makes it such that optional fields are truly optional and have a default value of `None` so that we can always upgrade schemas safely.

Test plan
=========

https://github.com/replit/ai-infra/pull/1309 wouldn't have required the new fields.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name="replit-river"
-version="0.2.2"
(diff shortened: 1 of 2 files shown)


cdmistman (approved):
🙌 thanks

PR #31, opened 2024-06-16
A few smol fixes

lhchavez:
Why
===

The Python implementation of River probably has a few bugs, since we see a large number of invariant violation errors.

What changed
============

This change:

* Stops a few resource leaks: the heartbeat tasks could get into a state where they never terminated, so they would consume a task forever.
* Stops a couple of TOCTOU races where an `await` point could have gotten in the way of a test-and-replace operation (which thanks to the GIL should be "atomic", except that pesky `await` point got in the way).
* Added a comment clarifying the transitions between the session states.
* Deindented a smol block for clarity.

Test plan
=========

Hopefully https://app.datadoghq.com/logs?query=%40river.tags%3Ainvariant-violation%20%40riverTarget%3Aai-chat%20&agg_m=count&agg_m_source=base&agg_q=service&agg_q_source=base&agg_t=count&cols=status_line%2Cmatches%2Cvolume%2Cservice%2Cmessage&fromUser=true&refresh_mode=sliding&storage=hot&stream_sort=time%2Cdesc&top_n=10&top_o=top&viz=pattern&x_missing=true&from_ts=1718406678000&to_ts=1718493078000&live=true occur less frequently.

### DIFF
--- a/replit_river/session.py
+++ b/replit_river/session.py
@@ -34,6 +34,11 @@ from .rpc import (
 
 
 class SessionState(enum.Enum):
+    """The state a session can be in.
+
+    Can only transition from ACTIVE to CLOSING to CLOSED.
+    """
+
     ACTIVE = 0
     CLOSING = 1
     CLOSED = 2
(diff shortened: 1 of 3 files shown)


replit_river/session.py
```diff
@@ -228,9 +237,15 @@ async def _check_to_close_session(self) -> None:
             await asyncio.sleep(
                 self._transport_options.close_session_check_interval_ms / 1000
             )
+            if self._state != SessionState.ACTIVE:
+                # already closing
+                return
+            # calculate the value now before comparing it so that there are no
```
zhenthebuilder, line 243: Trying to understand this better, Why would this be different before or after
```
if not self._close_session_after_time_secs:
  continue
```?

lhchavez: ah we don't want to have any `await` in the middle of doing a check / operation on the member field

zhenthebuilder: Ah I see your point, that make sense!
Thanks for the clarification, out of curiousity, how did you find this bug?

zhenthebuilder (approved):

PR #32, opened 2024-06-18
Proactively tell the server about our expected state

lhchavez:
Why
===

We have a few cases where the server recreates a connection that was meant to be a reconnect.

What changed
============

This change brings the Python implementation in line with the TypeScript one so that the server can proactively tell clients to recreate their sessions. See https://github.com/replit/river/pull/212 for reference

Test plan
=========

Logs like https://app.datadoghq.com/logs?query=%40replid%3Ae2698176-9a19-4058-8619-7f5793b02cde&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice%2C%40river.sessionId&event=AgAAAZAssKXztskj6QAAAAAAAAAYAAAAAEFaQXNzS2NBQUFBR1hKdG9Bd1NLLVFBbwAAACQAAAAAMDE5MDJjYjItYzE2Ny00YmZiLThjMjctNzdkNjI5Y2NkY2Vj&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=desc&viz=stream&from_ts=1718734624299&to_ts=1718737623927&live=false should completely go away

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name="replit-river"
-version="0.2.6"
+version="0.2.7"
 description="Replit river toolkit for Python"
 authors = ["Replit <eng@replit.com>"]
 license = "LICENSE"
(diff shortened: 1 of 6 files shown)


zhenthebuilder:
This PR looks good to me, should we also add logic in python server to handle expectedSessionState?

zhenthebuilder (approved):

lhchavez:
> This PR looks good to me, should we also add logic in python server to handle expectedSessionState?

oh yeah, brb

PR #34, opened 2024-06-28
Fix Type.any types issue with Codegen

vaikmr:
Why
===

`Type.any`  is codegened without a type. 

>data (and a few other fields) that are Type.Any are codegenned without a type field https://github.com/replit/ai-infra/pull/1423/files#diff-0dbf721bd6de1d6cb202caf7bd7a1412611f6680972f4748aad5a2e3af623a28R506
We then throw trying to codgen, likely right here https://github.com/replit/river-python/blob/main/replit_river/codegen/client.py#L95
[client.py](https://github.com/replit/river-python/blob/main/replit_river/codegen/client.py)

https://replit.slack.com/archives/C074CCJSS0M/p1719530737158639?thread_ts=1719528490.627619&cid=C074CCJSS0M

What changed
============

Fix the above issue.

Test plan
=========

Generate a client codegen from schema with no type defined.

### DIFF
--- a/replit_river/codegen/client.py
+++ b/replit_river/codegen/client.py
@@ -59,7 +59,7 @@ def encode_type(
     if isinstance(type, RiverConcreteType):
         if type.type is None:
             # Handle the case where type is not specified
-            return ("Any", ())  # or some other default behavior
+            return ("Any", ())
         if type.type == "string":
             if type.const:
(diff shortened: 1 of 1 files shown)


lhchavez (approved):
:ez:

PR #39, opened 2024-07-09
Prevent reconnects

lhchavez:
Why
===

We have some problems with reconnects, and it's probably better to completely disable transparent reconnects until we have that sorted out.

What changed
============

This change now proactively closes and deletes any outstanding sessions when we detect that the WebSocket transport was closed.

Test plan
=========

@cdmistman doesn't get into terrible states.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name="replit-river"
-version="0.2.18"
+version="0.2.19"
 description="Replit river toolkit for Python"
 authors = ["Replit <eng@replit.com>"]
 license = "LICENSE"
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -150,12 +150,26 @@ class ClientTransport(Transport):
             is_server=False,
             handlers={},
             close_session_callback=self._delete_session,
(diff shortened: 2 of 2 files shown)


jackyzha0:
is this safe to do? i don't remember if session closes notify procedures (i.e. what happens if we are mid-stream and this happens)

lhchavez:
> is this safe to do? i don't remember if session closes notify procedures (i.e. what happens if we are mid-stream and this happens)

session close _will_ close all streams https://github.com/replit/river-python/blob/main/replit_river/session.py#L506

jackyzha0 (approved):
i have a pr that should fix this, do we still want to merge this?

lhchavez:
if there's another PR that fixes it, we can land it instead.

cdmistman (approved):
i don't have the error about server-unknown sessions anymore!

lhchavez:
sorry, [this keeps happening](https://app.datadoghq.com/logs?query=%40replid%3A2be9b9ec-4a37-402d-a170-d4f6a4a8990c%20reconnect%20&agg_m=count&agg_m_source=base&agg_t=count&cols=core_host%2Ccore_service&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1721134374785&to_ts=1721220774785&live=true) , we need a big hammer.

PR #40, opened 2024-07-09
update river impl

jackyzha0:
Why
===

What changed
============

Test plan
=========

### DIFF
--- a/replit_river/rpc.py
+++ b/replit_river/rpc.py
@@ -44,27 +44,28 @@ ACK_BIT = 0x0001
 STREAM_OPEN_BIT = 0x0002
 STREAM_CLOSED_BIT = 0x0004
 
+# retriable codes
+# https://github.com/replit/river/blob/72e4fc8d02263e551d66ee5f0995707e8fa6cd1b/transport/message.ts#L83
+SESSION_MISMATCH_CODE = "SESSION_STATE_MISMATCH"
+
 # Equivalent of https://github.com/replit/river/blob/c1345f1ff6a17a841d4319fad5c153b5bda43827/transport/message.ts#L23-L33
-
-
 class ExpectedSessionState(BaseModel):
-    reconnect: bool
     nextExpectedSeq: int
+    nextSentSeq: Optional[int] = None
 
 
 class ControlMessageHandshakeRequest(BaseModel):
     type: Literal["HANDSHAKE_REQ"] = "HANDSHAKE_REQ"
     protocolVersion: str
     sessionId: str
-    expectedSessionState: Optional[ExpectedSessionState] = None
+    expectedSessionState: ExpectedSessionState
     metadata: Optional[Any] = None
 
 
 class HandShakeStatus(BaseModel):
     ok: bool
-    sessionId: Optional[str] = None
-    # Reason for failure
     reason: Optional[str] = None
+    code: Optional[str] = None
 
 
 class ControlMessageHandshakeResponse(BaseModel):
--- a/replit_river/server_transport.py
+++ b/replit_river/server_transport.py
@@ -57,20 +57,18 @@ class ServerTransport(Transport):
             to_id = msg.from_
             session_id = handshake_response.status.sessionId
             if not session_id:
-                raise InvalidMessageException("No session id in handshake response")
-            advertised_session_id = handshake_request.sessionId
+                raise InvalidMessageException("No session id in handshake request")
             try:
                 return await self.get_or_create_session(
                     transport_id,
                     to_id,
                     session_id,
-                    advertised_session_id,
                     websocket,
                 )
             except Exception as e:
                 error_msg = (
                     "Error building sessions from handshake request : "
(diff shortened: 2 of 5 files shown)


jackyzha0:
typescript change for ref: https://github.com/replit/river/pull/220

jackyzha0:
image

https://github.com/replit/river-babel/pull/36 babel passes

replit_river/session.py
```diff
@@ -503,7 +508,7 @@ async def _remove_acked_messages_in_buffer(self) -> None:
     async def start_serve_responses(self) -> None:
         self._task_manager.create_task(self.serve())
 
-    async def close(self, is_unexpected_close: bool) -> None:
```
lhchavez, line 506: rofl, we've been providing this field for naught all this time??

replit_river/server_transport.py
```diff
@@ -148,30 +146,69 @@ async def _establish_handshake(
...
-            my_session_id = await self._get_or_create_session_id(
-                request_message.from_, handshake_request.sessionId
-            )
-        elif handshake_request.expectedSessionState.reconnect:
-            maybe_my_session_id = await self._get_existing_session_id(
-                request_message.from_,
-                handshake_request.sessionId,
-                handshake_request.expectedSessionState.nextExpectedSeq,
+
+        oldSession = self._sessions.get(request_message.from_, None)
```
lhchavez, line 150:
```suggestion
        old_session = self._sessions.get(request_message.from_, None)
```

python likes snake_case for locals. same everywhere a new local is created.

replit_river/rpc.py
```diff
@@ -44,27 +44,30 @@
 STREAM_OPEN_BIT = 0x0002
 STREAM_CLOSED_BIT = 0x0004
 
-# Equivalent of https://github.com/replit/river/blob/c1345f1ff6a17a841d4319fad5c153b5bda43827/transport/message.ts#L23-L33
+# retriable codes
```
lhchavez, line 47: we need to define what "retriable" means here. e.g. HTTP makes a distinction between "you can retry the exact same request a-la 5XX" and "you can retry but only if you modify your request a-la 400, so it's not _really_ retriable".

by that distinction, this is more 4XX, and therefore are non-retriable in spirit.

lhchavez (approved):

replit_river/session.py
jackyzha0: i think it did something at some point ???

replit_river/rpc.py
jackyzha0: hm yeah seems like theres 3

1. retry safely without doing anything (nothing seems to really fall under here)
2. retry with a new session (session mismatch)
3. pls don't retry (protocol mismatch, etc.)

we only seem to have 2 and 3 cases for handshake so calling it retriable might be fine?

lhchavez: we shouldn't give it an ambiguous descriptor: we don't have a constraint on number of words in a comment, so might as well spell out explicitly what this means from a semantic perspective!

replit_river/server_transport.py
```diff
@@ -148,30 +146,74 @@ async def _establish_handshake(
...
+            client_next_sent_seq = (
+                handshake_request.expectedSessionState.nextSentSeq or 0
             )
-            if maybe_my_session_id is None:
-                handshake_response = await self._send_handshake_response(
+            our_next_seq = await old_session.get_next_sent_seq()
+            our_ack = await old_session.get_next_expected_seq()
+
+            if client_next_sent_seq > our_ack:
+                message = (
```
zhenthebuilder, line 165: Should we close & delete the old session here?

replit_river/server_transport.py
```diff
@@ -148,30 +146,74 @@ async def _establish_handshake(
                 websocket,
             )
             raise InvalidMessageException("handshake request to wrong server")
-        if handshake_request.expectedSessionState is None:
-            # TODO: remove once we have upgraded all clients
-            my_session_id = await self._get_or_create_session_id(
-                request_message.from_, handshake_request.sessionId
+
+        old_session = self._sessions.get(request_message.from_, None)
```
zhenthebuilder, line 150: Not sure if we could run into a race condition here, case might be :
1. we got a old_session here
2. control was given out while doing `await old_session.get_next_sent_seq()`
3. old session was deleted
4. When the control is given back the session is already gone

We probably want to wrap these logic with the session lock

replit_river/server_transport.py
jackyzha0: no, we tell the client to go away in this case to start a new session

jackyzha0: we dont want to delete sessions if we dont need to and would rather wait for the session to gracefully die

replit_river/server_transport.py
jackyzha0: which lock do we acquire?

jackyzha0: is it sufficient to wrap with `async with old_session._ws_lock, old_session._state_lock:`

PR #44, opened 2024-07-12
correctly decode optional oneof

bradymadden97:
Why
===

Our proto to python codegen treats optional fields as "oneof" fields, which expects an wrapper object with a $kind field. This doesn't match what our proto to typebox codegen does.

What changed
============

Test plan
=========

I wrote a proto with some optional fields, including an optional object field
Screenshot 2024-07-12 at 12 47 58 PM

Generated the decoder correctly, without expecting the wrapper object
Screenshot 2024-07-12 at 12 47 53 PM

### DIFF
--- a/replit_river/codegen/server.py
+++ b/replit_river/codegen/server.py
@@ -95,23 +95,34 @@ def message_decoder(
             )
     # oneof fields.
     for index, oneof in enumerate(m.oneof_decl):
-        chunks.extend(
-            [
-                f"  _{oneof.name} = d.get('{to_camel_case(oneof.name)}', {{}})",
-                f"  if _{oneof.name}:",
-                f"    match _{oneof.name}.get('$kind', None):",
-            ]
-        )
-        for field in oneofs[index]:
-            chunks.append(
-                f"        case '{to_camel_case(field.name)}':",
+        for fieldIndex, field in enumerate(oneofs[index]):
+            accessor_key = (
+                "d"
+                if field.label == FieldDescriptor.LABEL_OPTIONAL
+                else f"_{oneof.name}"
             )
+
+            # don't wrap optional fields in additional object
+            if field.label == FieldDescriptor.LABEL_OPTIONAL:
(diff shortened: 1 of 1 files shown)


replit_river/codegen/server.py
```diff
@@ -95,45 +95,56 @@ def message_decoder(
...
+            accessor_key = (
+                "d"
+                if field.label == FieldDescriptor.LABEL_OPTIONAL
+                else f"_{oneof.name}"
             )
+
+            # don't wrap optional fields in additional object
+            if field.label == FieldDescriptor.LABEL_OPTIONAL:
+                chunks.append(f"  if d.get('{to_camel_case(field.name)}') is not None:")
+            else:
```
lhchavez, line 108: we don't do a lot of ternaries, and we already have a branch for it.

```suggestion
            accessor_key: string

            # don't wrap optional fields in additional object
            if field.label == FieldDescriptor.LABEL_OPTIONAL:
                accessor_key = "d"
                chunks.append(f"  if d.get('{to_camel_case(field.name)}') is not None:")
            else:          
                accessor_key = f"_{oneof.name}"
```

replit_river/codegen/server.py
```diff
@@ -95,45 +95,56 @@ def message_decoder(
...
+                if fieldIndex == 0:
+                    chunks.extend(
+                        [
+                            f"_{oneof.name} ="
+                            f"d.get('{to_camel_case(oneof.name)}', {{}})",
+                            f"  if _{oneof.name}:",
+                            f"    match _{oneof.name}.get('$kind', None):",
```
lhchavez, line 115:
```suggestion
                            f"{accessor_key} ="
                            f"d.get('{to_camel_case(oneof.name)}', {{}})",
                            f"  if {accessor_key}:",
                            f"    match {accessor_key}.get('$kind', None):",
```

lhchavez (approved):

replit_river/codegen/server.py
bradymadden97: oh nice didn't know the pythonic way of doing that. easy enough

PR #50, opened 2024-07-19
fix infinite task cancellation loop

jackyzha0:
Why
===

- if we ever disconnected, shit would be bad

What changed
============

- add specific `transparent_reconnect` option
- make `_delete_session` async because it needs to acquire a lock
- lock the things appropriately

Test plan
=========

### DIFF
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -110,20 +110,26 @@ class ClientTransport(Transport):
                     if not old_session
                     else old_session.session_id
                 )
-                handshake_request, handshake_response = await self._establish_handshake(
-                    self._transport_id,
-                    self._server_id,
-                    session_id,
-                    self._handshake_metadata,
-                    ws,
-                    old_session,
-                )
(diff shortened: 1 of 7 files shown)


lhchavez (commented):
so far this looks legit!

JimmyAustin (approved):

replit_river/client_transport.py
```diff
@@ -286,32 +291,35 @@ async def _establish_handshake(
...
             )
+
         logging.debug("river client waiting for handshake response")
+        startup_grace_sec = 60
         try:
+            # listen even before we send
             response_msg = await asyncio.wait_for(
-                self._get_handshake_response_msg(websocket),
-                timeout=self._transport_options.session_disconnect_grace_ms / 1000,
+                self._get_handshake_response_msg(websocket), startup_grace_sec
```
jackyzha0, line 303: not strictly session disconnect grace, we also need to account for container startup time

zhenthebuilder (approved):

PR #51, opened 2024-07-19
Also close the WebSocket when we close the session

lhchavez:
Why
===

When we're closing sessions, we do not close the WebSocket and instead kinda rely on the server kicking us out due to inactivity. This makes it racy for cases like connection retries (which we have disabled and instead close any previous sessions), where we can have two concurrent connections for the same session id. An invariant violation!

What changed
============

This change now also makes sure to close the WebSocket when the session is closed.

Test plan
=========

See fewer instances of double concurrent conns.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name="replit-river"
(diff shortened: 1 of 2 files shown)


lhchavez:
nvm, it was closed a few lines above.

PR #54, opened 2024-07-24
chore/GHA release infrastructure

blast-hardcheese:
Why
===

Similar to [replit-py](https://github.com/replit/replit-py) and [replit-object-storage](https://github.com/replit/replit-object-storage-python), we should have a Release-oriented release process. This gives increased visibility into how our libraries change from version to version, as well as offering consistency for maintainers releasing changes.

What changed
============

- Adding a [`release-drafter`](https://github.com/release-drafter/release-drafter) config. This must be done before adding the workflow in a second PR, since the workflow always accesses the default branch for its configuration.
- Adding a `python-publish` action that publishes an update any time a new release is cut

Test plan
=========

Once merged, it should be possible to trigger the workflow to publish a new version of the library via workflow_dispatch or creating a release

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,33 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'feature'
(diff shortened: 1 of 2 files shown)


lhchavez (approved):
hooray!

PR #55, opened 2024-07-24
chore/Adding release-drafter

blast-hardcheese:
Why
===

Follow-up to #54, enable `release-drafter`.

What changed
============

Adding the release-drafter action to consume the configuration and curate release drafts on every merge to the default branch.

Test plan
=========

When this PR gets merged, do we get a new release draft?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,17 @@ This includes the necessary machinery to act as a client or server for River:
 
 ## Publishing
 
-Make sure you have pypi token setup, easiest way is to set env var `POETRY_PYPI_TOKEN_PYPI`.
+### Release Drafts
+Pending releases are curated by [release-drafter/release-drafter](https://github.com/release-drafter/release-drafter) on the [Releases](https://github.com/replit/river-python/releases) page.
 
-- update version either manually or via `poetry version`
-- `poetry build`
-- `poetry publish`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -14,8 +14,17 @@ This includes the necessary machinery to act as a client or server for River:
...
+
+### PR Labeling
+
+PRs merged since the last release are considered, with the labels on those PRs used for release metadata. `feature`, `bug`, `chore`, and `dependencies` are used for categorization, `major`, `minor`, and `patch` are used to influence the next release's version.
+
+These labels can be altered after merge, re-trigger release-drafter to get it to regenerate the draft once you've curated the next release.
+
+### Triggering release
+
+The tag version is used to set the version during the build, the value in `pyproject.toml` is not expected to be kept up-to-date.
```
jackyzha0, line 30: hm I don't think its obvious to me how to actually trigger a release from this, is there a button somewhere?

the way ruspty does this is just releasing when a version is bumped in package.json, can we just do something like that?

blast-hardcheese: Currently there's no drafted release, but you'd see something like this:

image

where when you hit the green `Publish` on the bottom, the git tag is created with the correct version number and the `publish-python` GHA is kicked off to actually do the release.

Happy to update the verbiage if necessary

jackyzha0 (approved):

PR #57, opened 2024-07-30
Improve logging / exception hygiene

lhchavez:
Why
===

We currently have some places where exceptions are not chained using the `raise ... from` statement, but rather just the message is added to another exception. That's suboptimal because we lose a lot of context.

What changed
============

This change now tries to use `raise ... from` in more places. Also, all error logs that embed an exception into the log entry now use `logging.exception` (which is equivalent plus it tacks the exception into the `exc_info` field), and other logs that embed the exception message into the log use `exc_info` instead (either setting it to `True` if it is logging in an `except` block, or manually setting the
  `exc_info` to the `BaseException` instance). Also, it surfaces
  exception chains in the stringification for uncaught errors.

Test plan
=========

:eyes: on :dog:.

### DIFF
--- a/replit_river/error_schema.py
+++ b/replit_river/error_schema.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, List, Optional
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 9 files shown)


replit_river/error_schema.py
```diff
@@ -21,3 +21,24 @@ def __init__(self, code: str, message: str) -> None:
         self.code = code
         self.message = message
         super().__init__(f"Error in river, code: {code}, message: {message}")
+
+
+def stringify_exception(e: BaseException, limit: int = 10) -> str:
+    """Return a string representation of an Exception.
+
+    This is different from just calling str(e) because it will also show the chained exceptions as
```
jackyzha0, line 29: huge

jackyzha0: might be nice to do first 5 last 5 causes if its more than 10?

jackyzha0 (approved):

replit_river/error_schema.py
lhchavez: typically the topmost are kinda useless since they typically live deep inside the machinery of asyncio, lol

lhchavez: (by "topmost" i mean Python-style because "most recent call last")

PR #59, opened 2024-08-09
More stable codegen

lhchavez:
Why
===

Welp, a change in pid2 changed the meaning of one of our anyOf enums! We don't want that to keep happening in the future. The real fix would be to have pid2 emit the discriminated unions on its own, but this is a hotfix.

What changed
============

This change does two things:
* Does the barest minimum attempt at detecting when an anyOf is a oneOf in disguise, and renders the types in a way that includes the discriminator value so that it is stable and will not change even if the lexicographic order of the entries change. Barest minimum since there is ongoing work to improve this.
* Also adds the value of the discrimnator as default in the `Field` constructor if the discriminator was renamed (i.e. if it was `$kind`), because mypy cannot fathom you being able to specify a parameter called `$kind` to the constructor, or being able to set that to a `Literal[...]`. It's weirdness all around.

Test plan
=========

https://github.com/replit/ai-infra/pull/1844 is no longer angry.

### DIFF
--- a/replit_river/codegen/client.py
+++ b/replit_river/codegen/client.py
@@ -1,8 +1,12 @@
 import json
+import re
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
 
 import black
 from pydantic import BaseModel, Field, RootModel
+
+_NON_ALNUM_RE = re.compile(r"[^a-zA-Z0-9_]+")
+_LITERAL_RE = re.compile(r"^Literal\[(.+)\]$")
 
 
 class RiverConcreteType(BaseModel):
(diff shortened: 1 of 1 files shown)


replit_river/codegen/client.py
```diff
@@ -49,6 +53,59 @@ def encode_type(
...
+                        and name == discriminator_name
+                        and prop.const is not None
+                    ].pop()
+                    type_name, type_chunks = encode_type(
+                        t, f"{prefix}OneOf_{discriminator_value}", base_model
+                    )
+                    chunks.extend(type_chunks)
+                    one_of.append(type_name)
+                if discriminator_name == "$kind":
+                    discriminator_name = "kind"
```
masad-frost, line 105: Maybe we should reconsider these dollar signs to eliminate the extra logic needed for python. Plus "type" is a more idiomatic in typescript, and we're using isinstance/classes/nominal typing in python

masad-frost (approved):
Tough read (codegen code), but looks great, thank you.

replit_river/codegen/client.py
```diff
@@ -49,6 +53,59 @@ def encode_type(
...
+            and t.type == "object"
+            and t.properties
+            and (not t.patternProperties or "^(.*)$" not in t.patternProperties)
+            for t in type.anyOf
+        ):
+            # We have established that it is a union-of-objects. Now let's see if
+            # there is a discriminator field common among all options.
+            literal_fields = set[str]()
+            for i, t in enumerate(type.anyOf):
+                # Mypy already forgot this fact D:
```
blast-hardcheese, line 68: since it's a `List` you're always gonna bubble back up to the ascribed type, unfortunately. Maybe just break out an explicit binding with the known type, if you don't want to keep doing this dance.

replit_river/codegen/client.py
blast-hardcheese: it's really Python's fault for not permitting `$` as a valid identifier character.

Alternately config file.
Alternately alternately, generic heuristic to translate or remove invalid identifier characters, with the assertion that there's no name conflict introduced.

blast-hardcheese (approved):
Logically, it makes sense. I do wonder if `_LITERAL_RE` is a sign that we are starting to push the limits of not having a real AST to play around with.

replit_river/codegen/client.py
lhchavez: yeah, that's a good idea. did that and added a big comment about it.

replit_river/codegen/client.py
lhchavez: this ^ re: Python's fault.

fortunately the only place where it really matters is when building an object literal in Python. for that case, the change in L164 will address that.

lhchavez:
> Logically, it makes sense. I do wonder if `_LITERAL_RE` is a sign that we are starting to push the limits of not having a real AST to play around with.

either introducing an Intermediate Representation (the AST can serve dual purpose in that way) or augmenting the return type so that it can hold the knowledge of a literal without us having to stringify it and then unsteingify it later would make this better, yes.

blast-hardcheese:
This unfortunately introduced an issue when rendering something in the `input` position:

```typescript
  Type.Union([
    Type.Object({
      $kind: Type.Literal('sshToken'),
      nonce: Type.String(),
      token: Type.String(),
      SSHHostname: Type.String(),
    }),
    Type.Object({
      $kind: Type.Literal('sshToken'),
      nonce: Type.String(),
      error: Type.String(),
    }),
  ]),
```

both get normalized into the same. Raising here for visibility, I've got a patch in-progress but I need to step away for a couple of hours.

PR #62, opened 2024-08-09
Make client accept a function for websocket uri and hadnshake metadata

masad-frost:
Why
===

We're seeing stale tokens as part of the connection to pid2, we have no way of regenerating the token

What changed
============

Make client accept a function for websocket uri and hadnshake metadata

Test plan
=========

Updated tests, unfortunately no comprehensive suite here so we'll test this in our internal repo


JimmyAustin (approved):
LGTM - Thanks for this!

PR #63, opened 2024-08-09
Follow-up to #61

blast-hardcheese:
Why
===

I didn't realize we didn't have codegen execution tests, I'll be more careful in the future.

What changed
============

Fixed two runtime issues

Test plan
=========

Manual testing


blast-hardcheese:
I scrutinized the generated code more carefully, comparing against the previous codegen. Things look better:

image

masad-frost (approved):
> I didn't realize we didn't have codegen execution tests, I'll be more careful in the future.

fun

PR #64, opened 2024-08-14
Better fire-and-forget WS closure

lhchavez:
Why
===

This introduces more problems than what it solves, since now if the application wants to exit, it may orphan this and cause a lot of logspam.

What changed
============

This change now closes the websocket synchronously to avoid leaking tasks upon exit.

Test plan
=========

Maybe fewer warnings about cancelling pending tasks?

### DIFF
--- a/replit_river/websocket_wrapper.py
+++ b/replit_river/websocket_wrapper.py
@@ -4,6 +4,8 @@ import logging
 
 from websockets import WebSocketCommonProtocol
 
+from replit_river.task_manager import BackgroundTaskManager
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


replit_river/websocket_wrapper.py
```diff
@@ -28,8 +28,8 @@ async def close(self) -> None:
         async with self.ws_lock:
             if self.ws_state == WsState.OPEN:
                 self.ws_state = WsState.CLOSING
-                task = asyncio.create_task(self.ws.close())
-                task.add_done_callback(
-                    lambda _: logger.debug("old websocket %s closed.", self.ws.id)
-                )
+                try:
+                    await self.ws.close()
```
zhenthebuilder, line 32: The ws.close() could take a long time to finish, and that's why we did a fire-and-forget here. We probably need to manage the task correctly here instead of let it be blocking.

BTW I think we actually don't need the self.ws_lock in this wrapper now, the lock in session.py is already doing its job.
Also the is_open check is not gonna race with close().

lhchavez: let's try the coroutine idea that we discussed irl.

jackyzha0:
closing for staleness, reopen if we still need?

PR #66, opened 2024-08-15
bug/Type union ignore-arg

blast-hardcheese:
Why
===

Any place we might encounter a `Union[...]`, we need `# type: ignore[arg-type]` to avoid mypy failing to typecheck, due to https://github.com/pydantic/pydantic/pull/8923, acknowledging that while pyright seems to have good inference, mypy fails to typecheck.

To better understand the issue and determine if there were any workarounds, I created https://github.com/blast-hardcheese/mypy-union-bug with the intent to report upstream, before finding the ongoing discussion threads linked here and in comments.

What changed
============

### Main

- Reflow format strings to use multiline format strings and `indent(dedent(s, ...))`. This gives us more horizontal space instead of trying to use C-style line continuation to avoid the 88-character limit
- Adding `# type: ignore[arg-type]` to possible `Union` slots

### 🧹 
- Tracking type literals more closely
- Reducing nested format strings for clarity
- Bubble up string-formatted comma for intelligibility

Test plan
=========

Manually running against production schemas, gathering sources together for a sensible local test suite for codegen

### DIFF
--- a/replit_river/codegen/client.py
+++ b/replit_river/codegen/client.py
@@ -1,5 +1,6 @@
 import json
 import re
+from textwrap import dedent, indent
 from typing import Any, Dict, List, Literal, Optional, Sequence, Set, Tuple, Union
 
 import black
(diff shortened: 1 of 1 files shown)


masad-frost (approved):

replit_river/codegen/client.py
```diff
@@ -242,166 +246,222 @@ def generate_river_client_module(
...
-                        "      input,",
-                        f"      lambda x: TypeAdapter({input_type}).dump_python(x, ",
-                        "        by_alias=True,",
-                        "        exclude_none=True,",
-                        "      ),",
-                        f"     {parse_output_method}",
-                        f"     {parse_error_method}",
-                        "    )",
+                        indent(
+                            dedent(
```
masad-frost, line 282: might be worth a comment, i would've been confused if not for the pr description

replit_river/codegen/client.py
```diff
@@ -429,9 +489,11 @@ def schema_to_river_client_codegen(
...
-            black.format_str(
-                "\n".join(generate_river_client_module(client_name, schemas.root)),
-                mode=black.FileMode(string_normalization=False),
+        s = "\n".join(generate_river_client_module(client_name, schemas.root))
+        try:
+            f.write(
+                black.format_str(s, mode=black.FileMode(string_normalization=False))
             )
-        )
+        except:
```
masad-frost, line 497: should we log that we failed to format?

blast-hardcheese: The failure to format is handled by the subsequent `raise`, where we're actually failing to parse the generated Python. This is just to help folks debug instead of being left with an empty output file.

PR #69, opened 2024-08-16
feat/TypedDict method inputs

blast-hardcheese:
Why
===

The way we do modeling with pydantic is challenging, most notably due to leaking the generated class names into client code.

By using TypedDict for input types, any dict literal that typechecks is a valid input. Some attention needs to be given to renamed parameters (`$kind`?) but once we're fully migrated over we can even just use `$kind` as property names natively due to them being strings in a dictionary instead of properties on a class.

What changed
============

- Threaded through `typeddict_encoder` generators
- Moved `procedure.input` over to `base_model='TypedDict'`, which worked better than expected

Test plan
=========

Manual testing so far.


masad-frost:
@JimmyAustin had some concerns around using typedict, did you two chat?

blast-hardcheese:
@masad-frost we did! This just introduces it for `input` types, to start, and it should start showing some UX benefits without changing how we're parsing responses from the backend. Before I merge this I'll be adding some tests to compare round-tripping.

blast-hardcheese:
Oh, @JimmyAustin , I misread @jackyzha0 . Jacky and I agree, Luis and I agree, happy to chat with James if there are concerns.

masad-frost (approved):
It's really hard to review earnestly, and probably not easy to write either. The approach makes sense

blast-hardcheese:
Soft-blocked by https://github.com/pydantic/pydantic/issues/10189

I'd rather not drop such a large change without a proper release, and we can't release to pypi currently due to having to depend on the git repo tip of pydantic to work around a bug

PR #72, opened 2024-08-20
feat/typed metadata

blast-hardcheese:
Why
===

From a discussion with Jacky, egged on by the revert of #62, we should know the schema of metadata that we're threading through.

What changed
============

`river.Client` is now bound by `A`, the schema of the structure passed in to `handshake_metadata`. If this agrees with the generated client's expectations, we're good to go.

Test plan
=========

Manual testing

### DIFF
--- a/replit_river/client.py
+++ b/replit_river/client.py
@@ -15,21 +15,21 @@ from .rpc import (
 logger = logging.getLogger(__name__)
 
 
-A = TypeVar("A")
+HandshakeType = TypeVar("HandshakeType")
 
 
-class Client(Generic[A]):
+class Client(Generic[HandshakeType]):
     def __init__(
         self,
         websocket_uri: str,
         client_id: str,
         server_id: str,
         transport_options: TransportOptions,
-        handshake_metadata: Optional[A] = None,
+        handshake_metadata: Optional[HandshakeType] = None,
     ) -> None:
         self._client_id = client_id
         self._server_id = server_id
-        self._transport = ClientTransport[A](
+        self._transport = ClientTransport[HandshakeType](
             websocket_uri=websocket_uri,
             client_id=client_id,
             server_id=server_id,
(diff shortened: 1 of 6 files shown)


replit_river/client.py
```diff
@@ -15,14 +15,19 @@
 logger = logging.getLogger(__name__)
 
 
-class Client:
+A = TypeVar("A")
```
jackyzha0, line 18: is it normal to have this be a one-letter generic? i would prefer something with more detail (e.g. `HandshakeMeta`) but in the case that this is just convention, a comment here would be nice

replit_river/server_transport.py
```diff
@@ -113,11 +113,11 @@ async def websocket_closed_callback() -> None:
     async def _establish_handshake(
         self, request_message: TransportMessage, websocket: WebSocketCommonProtocol
     ) -> Tuple[
-        ControlMessageHandshakeRequest,
+        ControlMessageHandshakeRequest[Any],
```
jackyzha0, line 116: `RiverConcreteType` instead?

jackyzha0 (approved):

replit_river/server_transport.py
blast-hardcheese: This is in the runtime, not the codegen, this is just codifying the previous behavior so none of the consumers break

jackyzha0: gotcha, cool w me

replit_river/client.py
blast-hardcheese: It's not super common since having a real name for a generic can often get confused with a class name or a real definition in some capacity. I'm sympathetic though, I'll give it a shot.

blast-hardcheese: Settled on `HandshakeType`, as to avoid any sort of overlap with real classes.

PR #76, opened 2024-08-29
Add RiverServiceException

JimmyAustin:
Why
===

Sometimes River has an issue, sometimes the service running River has an issue. We should split those out so that people know which is which.
 
What changed
============

Added `RiverServiceException` to differentiate the two. `RiverServiceException` also includes extra fields to include the service and procedure.

Test plan
=========

CI/CD, this change is backwards compatible due to RiverServiceException being a subclass of RiverException.

### DIFF
--- a/replit_river/error_schema.py
+++ b/replit_river/error_schema.py
@@ -7,6 +7,16 @@ ERROR_HANDSHAKE = "handshake_failed"
 ERROR_SESSION = "session_error"
 
 
+# UNCAUGHT_ERROR_ERROR_CODE is the code that is used when an error is thrown
+# inside a procedure handler that's not required.
+UNCAUGHT_ERROR_ERROR_CODE = 'UNCAUGHT_ERROR'
+
+# INVALID_REQUEST_ERROR_CODE is the code used when a client's request is invalid.
+INVALID_REQUEST_ERROR_CODE = 'INVALID_REQUEST'
+
+# CANCEL_ERROR_CODE is the code used when either server or client cancels the stream.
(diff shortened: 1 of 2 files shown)


zhenthebuilder (approved):

replit_river/error_schema.py
```diff
@@ -22,6 +22,25 @@ def __init__(self, code: str, message: str) -> None:
         self.message = message
         super().__init__(f"Error in river, code: {code}, message: {message}")
 
+class RiverServiceException(Exception):
```
zhenthebuilder, line 25: Inherit from RiverException

jackyzha0 (approved):
lgtm, do we know if theres info at this level about whether the channel is closed because of the connection dying or the other side telling us to close? would be helpful

in ts, we distinguish between
- session drop causing the stream to be closed https://github.com/replit/river/blob/bf6dd5477b2b945be7605cc6f181a67293a52ecb/router/errors.ts#L64
- procedure code threw https://github.com/replit/river/blob/bf6dd5477b2b945be7605cc6f181a67293a52ecb/router/errors.ts#L60
- server told us we made a bad request https://github.com/replit/river/blob/bf6dd5477b2b945be7605cc6f181a67293a52ecb/router/errors.ts#L68
- server proactively cancelled our request https://github.com/replit/river/blob/bf6dd5477b2b945be7605cc6f181a67293a52ecb/router/errors.ts#L72

we _probaby_ dont need that granularity here but would be nice to distinguish between 1 and 2/3/4

PR #77, opened 2024-08-30
Configure handshake timeout separately from session disconnect grace period

cbrewster:
Why
===

~We were magically computing a handshake timeout by dividing the session disconnect by 1000, which in production lead to a handshake timeout of 15ms (as session disconnect grace period is set to 15sec), which is pretty optimistic since RTT can be on the order of 100s of ms sometimes.~

Instead of computing this magically, let's allow it be configured explicitly and set it to a default of 10s which matches what we have set in repl-it-web.

What changed
============

- Add `handshake_timeout_ms` config with a default of 10s

Test plan
=========

- Ideally we see less handshake timeout errors in prod

### DIFF
--- a/replit_river/server.py
+++ b/replit_river/server.py
@@ -69,10 +69,11 @@ class Server(object):
         logger.debug(
             "River server started establishing session with ws: %s", websocket.id
         )
-        grace_ms = self._transport_options.session_disconnect_grace_ms / 1000
+        grace_ms = self._transport_options.handshake_timeout_ms
         try:
             session = await asyncio.wait_for(
-                self._handshake_to_get_session(websocket), grace_ms
+                self._handshake_to_get_session(websocket),
+                grace_ms / 1000,  # wait_for unit is seconds
             )
             if not session:
                 return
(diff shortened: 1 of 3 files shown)


cbrewster:

* main:
  * **PR #77** 👈

jackyzha0 (approved):

lhchavez (approved):
discussed offline about the possibility of adding a test.

lhchavez:
for reference, us-east4 <-> us-central1 latency is 28ms median https://console.cloud.google.com/net-intelligence/performance/global-dashboard?project=replit-web&pageState=(%22performanceDashboardTraffic%22:(%22v%22:%22internal%22),%22performanceDashboardMetric%22:(%22v%22:%22latency%22),%22pdNetworkTierState%22:(%22n%22:%22NETWORK_TIER_UNSPECIFIED%22),%22performanceDashboardRegions%22:(%22regions%22:%22%5B%5C%22us-central1%5C%22,%5C%22us-east4%5C%22%5D%22)) lol

cbrewster:
> discussed offline about the possibility of adding a test.

added a test

cbrewster:
Turns out dividing by 1000 was to convert from ms to sec, not to magically pick a good value here. But still this PR makes it more explicit and adds a test so worth landing.

PR #78, opened 2024-09-02
bug/conditionally enable typeddicts

blast-hardcheese:
Why
===

Unfortunately switching over to `TypedDicts` turned out to be overly optimistic without any sort of test suite to keep things in parity. For now, I'll make the codegen require explicit opt-in to not break clients.

What changed
============

- Conditionally opt-in to TypedDict inputs
- Bugfix: Encode arrays using the inner encoder
- Optional `Field(...`'s are invalid in pydantic, special case that.

Test plan
=========

Manual testing for now

### DIFF
--- a/replit_river/codegen/client.py
+++ b/replit_river/codegen/client.py
@@ -421,6 +421,7 @@ def encode_type(
 def generate_river_client_module(
     client_name: str,
     schema_root: RiverSchema,
+    typed_dict_inputs: bool,
 ) -> Sequence[str]:
     chunks: List[str] = [
         dedent(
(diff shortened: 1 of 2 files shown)


replit_river/codegen/client.py
```diff
@@ -304,11 +304,11 @@ def extract_props(tpe: RiverType) -> list[dict[str, RiverType]]:
             typeddict_encoder.append("None")
             return ("None", ())
         if type.type == "Date":
-            # typeddict_encoder.append("TODO")
+            typeddict_encoder.append("TODO: dstewart")
```
ryantm, line 307: What this todo about?

ryantm (commented):
It looks pretty straightforward, but what's up with those TODOs?

ryantm (approved):

PR #81, opened 2024-09-05
clear session if the session is closed to avoid using closed sessions

jackyzha0:
Why
===

when we get a `SESSION_STATE_MISMATCH`, we close the old session and delete it, but the recreate fn still has a handle to it

What changed
============

if the retry loop detects the old session to be closed, stop using it

Test plan
=========

### DIFF
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -112,7 +112,7 @@ class ClientTransport(Transport, Generic[HandshakeType]):
             rate_limit.consume_budget(client_id)
 
             # if the session is closed, we shouldn't use it
-            if not old_session or not await old_session.is_session_open():
+            if old_session and not await old_session.is_session_open():
                 old_session = None
 
             try:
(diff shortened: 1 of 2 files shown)


replit_river/client_transport.py
```diff
@@ -102,13 +102,19 @@ async def _establish_new_connection(
...
         for i in range(max_retry):
             if i > 0:
                 logger.info(f"Retrying build handshake number {i} times")
             if not rate_limit.has_budget(client_id):
                 logger.debug("No retry budget for %s.", client_id)
                 break
             rate_limit.consume_budget(client_id)
+
+            # if the session is closed, we shouldn't use it
+            if not old_session or not await old_session.is_session_open():
```
lhchavez, line 115:
```suggestion
            # if the session exists but is closed, we shouldn't use it
            if old_session and not await old_session.is_session_open():
```

lhchavez (approved):

PR #83, opened 2024-09-11
retry robustness

jackyzha0:
Why + What changed
===

- add details about what failed where possible instead of just overriding the error with "Failed to create ws after retrying max number of times"
- construct `handshake_metadata` earlier so if the await fails it is before ws is established

the plan is to also stop proc level retries within ai-infra on handshake errors because otherwise we get an explosion of connection attempts:
- single failure (either due to pid2 crashing or other reasons)
- causes retry at the proc level
- each proc level failure will retry the underlying transport connection 5 times

Test plan
=========

### DIFF
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -104,12 +104,16 @@ class ClientTransport(Transport, Generic[HandshakeType]):
         client_id = self._client_id
         logger.info("Attempting to establish new ws connection")
 
+        last_error = None
         for i in range(max_retry):
             if i > 0:
                 logger.info(f"Retrying build handshake number {i} times")
             if not rate_limit.has_budget(client_id):
                 logger.debug("No retry budget for %s.", client_id)
-                break
+                raise RiverException(
+                    ERROR_HANDSHAKE, f"No retry budget for {client_id}"
+                )
+
             rate_limit.consume_budget(client_id)
 
             # if the session is closed, we shouldn't use it
@@ -118,14 +122,13 @@ class ClientTransport(Transport, Generic[HandshakeType]):
 
             try:
                 websocket_uri = await self._websocket_uri_factory()
+                handshake_metadata = await self._handshake_metadata_factory()
                 ws = await websockets.connect(websocket_uri)
                 session_id = (
                     self.generate_session_id()
                     if not old_session
                     else old_session.session_id
                 )
-
-                handshake_metadata = await self._handshake_metadata_factory()
 
                 try:
                     (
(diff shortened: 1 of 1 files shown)


zhenthebuilder (approved):

replit_river/client_transport.py
```diff
@@ -104,12 +104,16 @@ async def _establish_new_connection(
...
         logger.info("Attempting to establish new ws connection")
 
+        last_error = None
         for i in range(max_retry):
             if i > 0:
                 logger.info(f"Retrying build handshake number {i} times")
             if not rate_limit.has_budget(client_id):
                 logger.debug("No retry budget for %s.", client_id)
-                break
+                raise RiverException(
```
zhenthebuilder, line 113: Do you want to log the last error here?

jackyzha0: good call

replit_river/client_transport.py
```diff
@@ -104,12 +104,25 @@ async def _establish_new_connection(
         client_id = self._client_id
         logger.info("Attempting to establish new ws connection")
 
+        last_error = None
```
lhchavez, line 107:
```suggestion
        last_error: Optional[Exception] = None
```

replit_river/client_transport.py
```diff
@@ -104,12 +104,25 @@ async def _establish_new_connection(
...
                 logger.debug("No retry budget for %s.", client_id)
-                break
+                if last_error:
+                    raise RiverException(
+                        ERROR_HANDSHAKE,
+                        (
+                            f"No retry budget for {client_id} after "
+                            f"last failure: {str(last_error)}"
+                        ),
+                    )
```
lhchavez, line 120:
```suggestion
                    raise RiverException(
                        ERROR_HANDSHAKE,
                        f"No retry budget for {client_id}"
                    ) from last_error
```

it's better to rely on Python to do the concatenation for us. that way we also preserve the OG stack :massage:

replit_river/client_transport.py
```diff
@@ -144,15 +156,23 @@ async def _establish_new_connection(
...
-            except Exception:
+            except Exception as e:
+                last_error = e
                 backoff_time = rate_limit.get_backoff_ms(client_id)
                 logger.exception(
-                    f"Error connecting, retrying with {backoff_time}ms backoff"
+                    (
+                        f"Error connecting: {str(e)}, "
+                        f"retrying with {backoff_time}ms backoff"
+                    )
```
lhchavez, line 166: do we _really_ need this? `logger.exception` should already add `e` into the message.

replit_river/client_transport.py
```diff
@@ -144,15 +156,23 @@ async def _establish_new_connection(
...
                 await asyncio.sleep(backoff_time / 1000)
+
         raise RiverException(
             ERROR_HANDSHAKE,
-            "Failed to create ws after retrying max number of times",
+            (
+                f"Failed to create ws after retrying {max_retry} number of times: "
+                f"{str(last_error)}"
+            ),
         )
```
lhchavez, line 176:
```suggestion
        raise RiverException(
            ERROR_HANDSHAKE,
            "Failed to create ws after retrying max number of times",
        ) from last_error
```

replit_river/client_transport.py
```diff
@@ -118,15 +131,14 @@ async def _establish_new_connection(
 
             try:
                 websocket_uri = await self._websocket_uri_factory()
+                handshake_metadata = await self._handshake_metadata_factory()
```
lhchavez, line 134: it'd be fun to have a callback thing that gets both (in another PR)

lhchavez (approved):

masad-frost (commented):
Noice

PR #85, opened 2024-09-14
Move ServerTransport specific methods from Transport to ServerTransport

masad-frost:
Why
===

_Describe what prompted you to make this change, link relevant resources: Linear issues, Slack discussions, etc._

What changed
============

_Describe what changed to a level of detail that someone with no context with your PR could be able to review it_

Test plan
=========

_Describe what you did to test this change to a level of detail that allows your reviewer to test it_

PR #84, opened 2024-09-14
Add a couple of environment setup files

masad-frost:
Why
===

Ez switching environments

What changed
============

Added .python-version and set virtualenv to in-project 

Test plan
=========

- pyenv works
- creating virtual env creates it in project

PR #86, opened 2024-09-14
Add ensure_connected method to client_transport

masad-frost:
Why
===

There's currently no way of making sure we're connected other than sending an RPC. This poses issues where we have errors propagating much later than they should be.

What changed
============

Add method to allow us to check if we're connected and error as early as possible

### DIFF
--- a/replit_river/client.py
+++ b/replit_river/client.py
@@ -42,6 +42,9 @@ class Client(Generic[HandshakeType]):
         await self._transport.close()
         logger.info(f"river client {self._client_id} closed")
 
+    async def ensure_connected(self) -> None:
+        await self._transport.get_or_create_session()
+
     async def send_rpc(
         self,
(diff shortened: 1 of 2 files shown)


masad-frost:

* main:
  * **PR #84**
    * **PR #85**
      * **PR #86** 👈

blast-hardcheese (approved):
not to bikeshed, but would a better encoding be an async context manager? Even if we don't close it at the end, it would more accurately represent what we're trying to do instead of a nullary unit function

masad-frost:
Yeah, I think that's the best/most pythonic way to do it. I'm just trying to fit it in the existing structure we have, it's probably worth fixing down the line. I'm trying to get connection errors to bubble up as early as possible and we don't have a way to do that right now (other than triggering an RPC)

PR #87, opened 2024-09-23
Combine uri and metadata factory

masad-frost:
Why
===

It's unnecessary to have two functions. With the way our metadata and websocket uri are setup, it's just added complexity

What changed
============

- Make `handshake_metadata_factory` and `websocket_uri_factory` a single `uri_and_metadata_factory` that grabs both. 
- Returns a new TypeDict `UriAndMetadata`
- Rename `HandshakeType` to `HandshakeMetadataType`
- `HandshakeMetadataType` `TypeVar` is reused (not sure if i'm doing some python no-no here). Also moved it to the options module to avoid circular imports

### DIFF
--- a/replit_river/client_transport.py
+++ b/replit_river/client_transport.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 from collections.abc import Awaitable, Callable
-from typing import Generic, Optional, Tuple, TypeVar
+from typing import Generic, Optional, Tuple
 
 import websockets
 from pydantic import ValidationError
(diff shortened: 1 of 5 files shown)


masad-frost:

* main:
  * **PR #85**
    * **PR #86**
      * **PR #87** 👈

replit_river/client_transport.py
```diff
@@ -45,27 +45,30 @@
 HandshakeType = TypeVar("HandshakeType")
 
 
+class UriAndMetadata(TypedDict, Generic[HandshakeType]):
```
masad-frost, line 48: thanks python 3.11, we can have TypeDicts with generics

jackyzha0 (approved):

PR #92, opened 2024-09-27
[flake] embrace our new nix overlords tighter

ryantm:
Why
===
* We didn't install python or poetry last time

What changed
===
* Add python 3.11 and poetry

Test plan
===
* flake built on linux and poetry was workin'


lhchavez (approved):
:people_hugging:

PR #95, opened 2024-09-28
feat/finishing typeddict inputs

blast-hardcheese:
Why
===

We got pretty close to having TypedDicts for river-python inputs before, but had to roll back due to a protocol mismatch.

Trying again, and also adding some tests to confirm that at the very least the Pydantic models can decode was was encoded by the TypedDict encoders. It's not a perfect science, but it should be good enough to start building more confidence as we make additional progress.

### The reason for "janky" tests

There's a bit of a chicken-and-egg situation when trying to test code generation at runtime.

We have three options:
- write pytest handlers where each invocation runs the codegen with a temp target (like the shell script does here), writes a static file for each text into that directory, then executes a new python into that directory. The challenge with this is that it would suck to write or maintain.
- write pytest handlers which runs the codegen with unique module name targets (like `gen1`, `gen2`, `gen3`, one for each codegen run necessary) and carefully juggle the imports to make sure we don't try to import something that's not there yet. This _might_ be the best option, but I'm not convinced about the ergonomics at the moment. It might be OK though, with highly targeted `.gitignore`'s.
- maintain a bespoke test runner, optimize for writing and maintaining these tests, and just acknowledge that we are doing something obscure and difficult.

I definitely wrote the tests here in a way that would give some coverage and also provide confidence, while intentionally deferring the above decision so we can keep making progress. in the meantime.

What changed
============

- Added some janky tests for comparing the encoding of both models
- Fixed many bugs in the TypedDict codegen and encoders

Test plan
=========

```
$ bash scripts/parity.sh
Using /tmp/river-codegen-parity.bAZ
Starting...
Verified
```

### DIFF
--- a/replit_river/codegen/client.py
+++ b/replit_river/codegen/client.py
@@ -273,44 +273,44 @@ def encode_type(
             # Handle the case where type is not specified
             typeddict_encoder.append("x")
             return ("Any", ())
-        if type.type == "string":
+        elif type.type == "string":
             if type.const:
                 typeddict_encoder.append(f"'{type.const}'")
                 return (f"Literal['{type.const}']", ())
             else:
                 typeddict_encoder.append("x")
                 return ("str", ())
-        if type.type == "Uint8Array":
+        elif type.type == "Uint8Array":
             typeddict_encoder.append("x.decode()")
             return ("bytes", ())
-        if type.type == "number":
+        elif type.type == "number":
             if type.const is not None:
                 # enums are represented as const number in the schema
                 typeddict_encoder.append(f"{type.const}")
                 return (f"Literal[{type.const}]", ())
             typeddict_encoder.append("x")
             return ("float", ())
-        if type.type == "integer":
+        elif type.type == "integer":
             if type.const is not None:
                 # enums are represented as const number in the schema
                 typeddict_encoder.append(f"{type.const}")
                 return (f"Literal[{type.const}]", ())
             typeddict_encoder.append("x")
             return ("int", ())
-        if type.type == "boolean":
+        elif type.type == "boolean":
             typeddict_encoder.append("x")
             return ("bool", ())
-        if type.type == "null":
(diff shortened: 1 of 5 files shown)


blast-hardcheese:
grr, it doesn't like the janky tests.

blast-hardcheese:
@ryantm I edited the description to explain the odd testing story here

replit_river/codegen/client.py
```diff
@@ -353,18 +359,35 @@ def extract_props(tpe: RiverType) -> list[dict[str, RiverType]]:
...
+                                typeddict_encoder.append(
+                                    f"""[
+                                        encode_{_inner_type_name}(y)
+                                        for y in x['{name}']
+                                        ]"""
+                                )
                         else:
-                            typeddict_encoder.append(f"x['{safe_name}']")
+                            if name in prop.required:
+                                typeddict_encoder.append(f"x['{safe_name}']")
```
ryantm, line 388: You bind name here, but don't use it. Is that intended (maybe its just checking its there)?

scripts/parity/check_parity.py
```diff
@@ -0,0 +1,156 @@
...
+    Literal[18],
+    Literal[19],
+    Literal[20],
+    Literal[21],
+    Literal[22],
+    Literal[23],
+    Literal[24],
+    Literal[25],
+    Literal[26],
+    None,
```
ryantm, line 86: :laughing:

ryantm (approved):

replit_river/codegen/client.py
blast-hardcheese: not actually a binding, checking to see if `name` is in the list of required

scripts/parity/check_parity.py
blast-hardcheese: alas, python is not yet a dependently typed language 😉

PR #97, opened 2024-10-08
Add backpressure to send buffer

cbrewster:
Why
===

Currently the send buffer has no backpressure, so if a user of the river client sends too many messages too quickly, the send buffer will overflow and the client will error out.

Rather than relying on users to rate limit themselves, we introduce backpressure so client naturally will only send messages as fast as the river client/server is able to consume them.

What changed
============

- Add a condvar to the message buffer to allow waiting for buffer space to come available when the buffer is full
- Add a closed bit to the message buffer so when a session is shut down, we have a way of unblocking all the futures that are stuck waiting for space in the buffer
- Close the message buffer when the session is closed

Test plan
=========

- Added a test which sends `2 * MAX_MESSAGE_BUFFER_SIZE` messages to a river server. Previously this resulted in an error.
- Added some unit tests for the `MessageBuffer` to ensure backpressure works and closing works

### DIFF
--- a/replit_river/message_buffer.py
+++ b/replit_river/message_buffer.py
@@ -8,6 +8,10 @@ from replit_river.transport_options import MAX_MESSAGE_BUFFER_SIZE
 logger = logging.getLogger(__name__)
 
 
+class MessageBufferClosedError(BaseException):
+    """Raised when a message buffer is closed and is not accepting new messages."""
+
+
 class MessageBuffer:
     """A buffer to store messages and support current updates"""
 
@@ -15,6 +19,8 @@ class MessageBuffer:
         self.max_size = max_num_messages
         self.buffer: list[TransportMessage] = []
         self._lock = asyncio.Lock()
+        self._space_available_cond = asyncio.Condition(lock=self._lock)
+        self._closed = False
 
     async def empty(self) -> bool:
         """Check if the buffer is empty"""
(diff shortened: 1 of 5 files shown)


cbrewster:

* main:
  * **PR #97** 👈

replit_river/message_buffer.py
```diff
@@ -23,10 +29,12 @@ async def empty(self) -> bool:
 
     async def put(self, message: TransportMessage) -> None:
         """Add a message to the buffer"""
-        async with self._lock:
-            if len(self.buffer) >= self.max_size:
-                logger.error("Buffer is full, dropping message")
-                raise ValueError("Buffer is full")
```
lhchavez, line 29: do we need to do something on the callsite to not care about this anymore?

tests/test_message_buffer.py
```diff
@@ -0,0 +1,64 @@
...
+        from_="test",
+        to="test",
+        streamId="test",
+        controlFlags=0,
+        payload=0,
+        model_config={},
+    )
+
+
+async def test_message_buffer_backpressure():
```
lhchavez, line 23: based mypy.

replit_river/message_buffer.py
```diff
@@ -8,13 +8,19 @@
...
 
 
+class MessageBufferClosedError(BaseException):
+    """Raised when a message buffer is closed and is not accepting new messages."""
+
+
 class MessageBuffer:
     """A buffer to store messages and support current updates"""
 
     def __init__(self, max_num_messages: int = MAX_MESSAGE_BUFFER_SIZE):
```
lhchavez, line 18: now that we have backpressure, it might be beneficial to reduce this a bit to prevent OOMs!

lhchavez (approved):
heck yeah!

replit_river/message_buffer.py
cbrewster: Agree, let's shrink to 128?

PR #98, opened 2024-10-10
[replit_river] close websocket right away whe closing

ryantm:
Why
===
* We were seeing lots of log spam at the end of test runs starting with

> Task was destroyed but it is pending!

What changed
===
* Instead of enqueuing a websocket close task while closing the websocket wrapper, immediately close the websocket

Test plan
===
* Run any e2e test in ai with this new version and see there is no longer a bunch of log spam


jackyzha0:
i believe we intentionally had this because the ws could take a while to close sometimes, ill wait for zhen to chime in

ryantm:
Okay, thanks for the context. I found another alternative solution!

PR #99, opened 2024-10-11
[replit_river] return cleanup task from client.disconnect()

ryantm:
Why
===
* The task created by the websocket wrapper was orphaned. Tasks need to be awaited somewhere, or you get errors like
```
RuntimeError: no running event loop
Task was destroyed but it is pending!
```
* Since it takes a while to finish, we don't want to wait for it in certain cases, so instead we'll return it as a cleanup task that the caller can await as appropriate.

What changed
===
* When the websocket close task is made, return it.
* At every level, return the task and combine it with other cleanup tasks as appropriate

Test plan
===
* The behavior shouldn't be different unless you await the cleanup function. If you do await it, you won't get a pending task exception when closing the event loop.

### DIFF
--- a/replit_river/client.py
+++ b/replit_river/client.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
 from typing import Any, Generic, Optional, Union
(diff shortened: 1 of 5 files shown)


masad-frost (approved):
Nice

PR #100, opened 2024-10-11
[websocket] use background tasks instead of cleanup function 

ryantm:
Why
===
* The cleanup function added a burden on the caller to make sure the
tasks were awaited.
* Instead, follow the advice of
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
and store them in the background properly

What changed
===
* Revert cleanup commit
* Add background websocket task storage

Test plan
===
* Ran a pytest, and didn't see "Pending task destroyed" exceptions

### DIFF
--- a/replit_river/websocket_wrapper.py
+++ b/replit_river/websocket_wrapper.py
@@ -5,6 +5,7 @@ import logging
 from websockets import WebSocketCommonProtocol
 
 logger = logging.getLogger(__name__)
+_background_tasks: set[asyncio.Task] = set()
 
 
 class WsState(enum.Enum):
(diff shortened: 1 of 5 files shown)


replit_river/websocket_wrapper.py
```diff
@@ -5,6 +5,7 @@
 from websockets import WebSocketCommonProtocol
 
 logger = logging.getLogger(__name__)
+_background_tasks: set[asyncio.Task] = set()
```
lhchavez, line 8: _fine_. don't love the idea of having another global, but the alternative is to re-add all the complexity. lgtm!

lhchavez (approved):

PR #101, opened 2024-10-24
[bug] Missed the "init" encoders

blast-hardcheese:
Why
===

- TypedDict attempts are failing because we're passing TypedDicts into pydantic. Fix this.

What changed
============

- Follow the input encoder pattern for now. Before generalizing I want to have `NewType`s to keep bindings straight

Test plan
=========

Manually linking the project and running codegen makes tests pass


ryantm (approved):
I don't fully get it, but doesn't look dangerous and we have a lot of type checking that will check this.

PR #104, opened 2024-11-04
[bug/client codegen] local_discriminator should be stable

blast-hardcheese:
What changed and Why
===

When there are multiple discriminator choices, we should just choose the first one.

Test plan
=========

Manual verification via repl


masad-frost (approved):
So this changes it from first-seen to alphabetical?

PR #106, opened 2024-11-06
feat/river schema v2 compatibility

blast-hardcheese:
Why
===

River Schema v2 compatibility so we can start getting rid of the compatibility layer.

What changed
============

Client
---

Significantly restructured the `river.v2.*`.

- Nearly all `asyncio.Lock`'s have been removed
- Background tasks are now cleaned up correctly, which should reduce the amount of noise around tasks being GC'd while they're still running
- Move websocket state ownership from Client transport down into Client Session. This significantly simplifies transparent reconnects, since we don't need to coordinate state shifting.
- There are four background tasks:
  - Heartbeat checker. This keeps track of whether we've gotten any heartbeats from the server recently.
  - "Close session" checker. This keeps track of whether we've exceeded the grace period for transparent reconnects.
  - Buffered message sender. One downside with the previous implementation was that we had [a race](https://us5.datadoghq.com/logs?query=%40river.tags%3Ainvariant-violation%20-%40river.extras.badASN%3Atrue%20service%3Apid2%20status%3Awarn%20%22received%20out%20of%20order%20msg%20closing%20connection%20got%20seq%22%20%22wanted%20seq%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1742751250088&to_ts=1743010450088&live=true) on sending messages over the websocket, since we had no coordination between calls to `send_message` when called concurrently, so we'd end up with `seq` bound and incremented but then messages would become out-of-order by the time the `ws.send` method returned. Now we just atomically append to a [`deque`](https://docs.python.org/3/library/collections.html#collections.deque) as we push onto the send buffer, and when we get a confirmation from the ws library we move the message over to an ack buffer (also a `deque`) waiting for incoming messages from the server to allow us to drain.

Codegen
---

- `init` is now required for all method types, `input` is now optional. This simplifies `rpc` and `stream` codegen.
- Fixed some bugs where `init` was being used in `input` position and vice versa.
- Swapping `f"'{foo}'"` encoding for `f"{repr(foo)}"`, giving greater safety post-generation (to avoid situations where `f"'{None}'"` gets rendered as `"None"`. `f"{repr(None)}"` would render as `None` which would fail typechecks.)

Test plan
=========

Manually ran codegen against v2 generated schemas, everything typechecks.


replit_river/codegen/client.py
```diff
@@ -209,8 +207,8 @@ def flatten_union(tpe: RiverType) -> list[RiverType]:
                         typeddict_encoder.append(f"encode_{type_name}(x)")
                     typeddict_encoder.append(
                         f"""
-                            if x['{discriminator_name}']
-                            == '{discriminator_value}'
+                            if x[{repr(discriminator_name)}]
+                            == {repr(discriminator_value)}
```
masad-frost, line 211: huh, how did this even work lol

replit_river/client_session.py
```diff
@@ -83,10 +83,10 @@ async def send_upload(
...
         service_name: str,
         procedure_name: str,
-        init: Optional[InitType],
-        request: AsyncIterable[RequestType],
-        init_serializer: Optional[Callable[[InitType], Any]],
-        request_serializer: Callable[[RequestType], Any],
+        init: InitType,
+        request: Optional[AsyncIterable[RequestType]],
+        init_serializer: Callable[[InitType], Any],
+        request_serializer: Optional[Callable[[RequestType], Any]],
```
masad-frost, line 89: Ideally the type is if `request` exists then `request_serializer` should not be optional.

replit_river/client_session.py
```diff
@@ -100,29 +100,26 @@ async def send_upload(
...
-                    payload=request_serializer(item),
-                )
+            await self.send_message(
+                stream_id=stream_id,
+                control_flags=STREAM_OPEN_BIT,
+                service_name=service_name,
+                procedure_name=procedure_name,
+                payload=init_serializer(init),
+            )
+            first_message = False
```
masad-frost, line 110: `first_message` should be no longer needed since we always send a `STREAM_OPEN_BIT` and we can remove the bit from `send_close_stream`

replit_river/client_session.py
```diff
@@ -100,29 +100,26 @@ async def send_upload(
...
+            if request is not None and request_serializer is not None:
+                # If this request is not closed and the session is killed, we should
+                # throw exception here
+                async for item in request:
+                    control_flags = 0
+                    await self.send_message(
+                        stream_id=stream_id,
+                        service_name=service_name,
+                        procedure_name=procedure_name,
+                        control_flags=control_flags,
```
masad-frost, line 120:
```suggestion
                    await self.send_message(
                        stream_id=stream_id,
                        service_name=service_name,
                        procedure_name=procedure_name,
                        control_flags=0,
```

replit_river/codegen/client.py
```diff
@@ -623,24 +635,20 @@ def __init__(self, client: river.Client[{handshake_type}]):
...
                       self,
-                      input: {input_type},
+                      init: {init_type},
                     ) -> {output_type}:
-                      {control_flow_keyword}await self.client.send_rpc(
-                        '{schema_name}',
-                        '{name}',
-                        input,
-                        {render_input_method},
+                      return await self.client.send_rpc(
```
masad-frost, line 647: all this stuff is hard to review (and I'm sure iterate on), we should set up some simple tests. Snapshot testing is probably good enough https://pypi.org/project/pytest-snapshot/.

I did this for river js https://github.com/replit/river/pull/261 using https://vitest.dev/guide/snapshot

replit_river/codegen/server.py
```diff
@@ -155,11 +155,11 @@ def message_encoder(
...
         f"def _{m.name}Encoder(",
         f"  e: {module_name}_pb2.{m.name}",
-        ") -> Dict[str, Any]:",
-        "  d: Dict[str, Any] = {}",
+        ") -> dict[str, Any]:",
+        "  d: dict[str, Any] = {}",
     ]
     # Non-oneof fields.
-    oneofs: DefaultDict[int, List[descriptor_pb2.FieldDescriptorProto]] = (
+    oneofs: DefaultDict[int, list[descriptor_pb2.FieldDescriptorProto]] = (
```
masad-frost, line 162: What's the difference?

I wonder if it's like the stupid typescript thing where `String` (object) is different from `string` (literal/primitive) :grimacing:

masad-frost (approved):
Do you wanna set up a different base branch in case we need to send a patch against main?

replit_river/codegen/client.py
blast-hardcheese: It's equivalent in the happy path but `repr` will surface bugs in the not happy path

replit_river/client_session.py
blast-hardcheese: Based on how the code worked previously, if we had an exception during sending the first message, then `STREAM_OPEN_BIT` was sent in the `send_close_stream` -- was that not intentional?

replit_river/codegen/client.py
blast-hardcheese: The best way I have found to iterate on this is to just run the codegen against different production schemas and verify that the set of changes is what I expect. I'm happy to wait until after I land the babel changes before merging this.

The codegen currently isn't compositional enough to make snapshot testing manageable. The more I work with it though I've been trying to come up with strategies to make it less imperative without increasing cognitive load too much, I'll land snapshot testing then.

replit_river/codegen/server.py
blast-hardcheese: Python 3.8 and below did not permit the builtin types to be subscripted, `typing.List` and `typing.Dict` were added as future-compatible shims during the migration. Python 3.8 is EOL as of the end of Spooky Season, so now we can 🚀.

replit_river/client_session.py
blast-hardcheese: Maybe it should be `request: Optional[({request_type}, {request_serializer_type})]`

blast-hardcheese:
> Do you wanna set up a different base branch in case we need to send a patch against main?

If we need to patch, we should do feature patches off previous release tags. Since we're reserving 1.x for GA, I can follow the 0.200.0 strategy in this repo for schema-v2 releases.

Doing that though would suggest that we do at least a halfway decent pass at the v2 server codegen as well though. Given what we discussed supporting v1/v2 using a naive shim doesn't seem too bad, probably can get that done this week.

Wdyt?

replit_river/codegen/client.py
blast-hardcheese: Snapshot tests are now in place, but only for v1 right now. I'll add some for v2.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1290 @@
...
+        # come, first served.
+        #
+        # If somebody else is already waiting or we've filled the buffer, we
+        # should get in line.
+        if (
+            self._queue_full_lock.locked()
+            or len(self._send_buffer) >= self._transport_options.buffer_size
+        ):
+            logger.debug("_send_message: queue full, waiting")
+            await self._queue_full_lock.acquire()
```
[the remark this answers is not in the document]
cbrewster: Ah I see, the problem with the old approach was out of order `seq` :thinking:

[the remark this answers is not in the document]
blast-hardcheese: Switching from `Lock` to `Condition` would let us `notify_all()` on close, good catch

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1290 @@
...
+        # Clear the session in transports
+        # This will get us GC'd, so this should be the last thing.
+        await self._close_session_callback(self)
+
+    def _start_buffered_message_sender(self) -> None:
+        def commit(msg: TransportMessage) -> None:
+            pending = self._send_buffer.popleft()
+            if msg.seq != pending.seq:
+                logger.error("Out of sequence error")
+            self._ack_buffer.append(pending)
```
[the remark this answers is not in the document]
blast-hardcheese: They're the head/tail of the same window of messages. `_send_buffer` is ones that we're trying to send, `_ack_buffer` are the ones we're waiting for the server to ack receipt of.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1290 @@
...
+    ) -> A:
+        """Sends an upload request to the server.
+
+        Expects the input and output be messages that will be msgpacked.
+        """
+
+        stream_id = nanoid.generate()
+        output: Channel[Any] = Channel(1)
+        self._streams[stream_id] = output
+        try:
```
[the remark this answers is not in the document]
blast-hardcheese: that must be what this is about https://github.com/replit/river-python/blob/a41caa21a9d624609c28d0713ce144d213088e5e/src/replit_river/v2/session.py#L672-L673. I can shore that up in this PR.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1290 @@
...
+                # during the cleanup procedure.
+                self._terminating_task = asyncio.create_task(self.close())
+
+        async def transition_connected(ws: ClientConnection) -> None:
+            self._state = SessionState.ACTIVE
+            self._ws = ws
+
+            # We're connected, wake everybody up
+            async with self._connection_condition:
+                self._connection_condition.notify_all()
```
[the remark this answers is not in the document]
blast-hardcheese: You were right! I was thinking in a threaded world, not an async one. `Event` makes much more sense if we can rely on our Task having exclusive control of the Session state.

I was able to remove all Locks and the Semaphore by switching to `Event`, ensuring they're `set()` during `close()` to wake up all emitters. Much simpler.

src/replit_river/v2/session.py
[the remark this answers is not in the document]
blast-hardcheese: Done!

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1290 @@
...
+            # This is safe because each individual function that is waiting on this
+            # function completeing already has a reference, so we'll last a few ticks
+            # before GC.
+            #
+            # Let's do our best to avoid clobbering other tasks by comparing the .name
+            current_task = asyncio.current_task()
+            if (
+                self._connecting_task
+                and current_task
+                and self._connecting_task.get_name() == current_task.get_name()
```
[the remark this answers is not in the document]
blast-hardcheese:
```python
import asyncio

y: asyncio.Task | None = None


async def bar() -> None:
    ours = asyncio.current_task()
    assert ours
    print("Same(bar)?", ours is y)

async def foo() -> None:
    ours = asyncio.current_task()
    assert ours
    print("Same(foo)?", ours is y)
    await bar()


async def main() -> None:
    global y
    x = asyncio.create_task(foo())
    x = asyncio.create_task(foo())
    y = x

    await x


asyncio.run(main())
```

Unable to repro in an isolated way:

```
Same(foo)? False
Same(bar)? False
Same(foo)? True
Same(bar)? True
```

I wonder if it was some trick of the light, or possibly an indication of poor state management during the refactor.

src/replit_river/codegen/client.py
```diff
@@ -798,10 +806,314 @@ def __init__(self, client: river.Client[{handshake_type}]):
...
+            _, init_type, render_init_method = init_meta
+            current_chunks.extend(
+                [
+                    reindent(
+                        "  ",
+                        f"""\
+        async def {name}(
+          self,
+          init: {render_type_expr(init_type)},
+        ) -> {  # TODO(dstewart) This should just be output_type
```
jackyzha0, line 991: hm why do we want output_type over output_or_error_type

src/replit_river/codegen/run.py
```diff
@@ -38,9 +39,27 @@ def main() -> None:
...
+    client.add_argument(
+        "--protocol-version",
+        help="Generate river v2 clients",
+        action="store",
+        default="v1.1",
+        choices=["v1.1", "v2.0"],
+    )
+    client.add_argument(
+        "--method-filter",
+        help="Only generate a subset of the specified methods",
```
jackyzha0, line 51: help string should probably include that this is a path (unless `type=pathlib.Path` already does that)

src/replit_river/common_session.py
```diff
@@ -29,107 +33,69 @@ async def __call__(
...
     """
 
-    ACTIVE = 0
-    CLOSING = 1
-    CLOSED = 2
+    PENDING = 0
+    CONNECTING = 1
+    ACTIVE = 2
+    CLOSING = 3
+    CLOSED = 4
```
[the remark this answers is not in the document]
jackyzha0: what does a PENDING state mean? is this equivalent to the NoConnection state in ts (maybe we should keep the terminology consistent)

src/replit_river/common_session.py
```diff
@@ -29,107 +33,69 @@ async def __call__(
...
+    commit: Callable[[TransportMessage], None],
 ) -> None:
-    logger.debug("Start heartbeat")
     while True:
-        await asyncio.sleep(heartbeat_ms / 1000)
-        state = get_state()
-        if state != SessionState.ACTIVE:
+        await message_enqueued.acquire()
+        while (ws := get_ws()) is None:
+            # Block until we have a handle
```
jackyzha0, line 66: what happens if we never reconnect?

src/replit_river/common_session.py
```diff
@@ -29,107 +33,69 @@ async def __call__(
...
-                payload={
-                    "ack": 0,
-                },
-                control_flags=ACK_BIT,
-                procedure_name=None,
-                service_name=None,
-                span=None,
+            async with connection_condition:
+                await connection_condition.wait()
+        if msg := get_next_pending():
```
jackyzha0, line 73: does this not busy loop if there are pending messages and a ws is open?

src/replit_river/common_session.py
```diff
@@ -29,107 +33,69 @@ async def __call__(
...
-            if increment_and_get_heartbeat_misses() > heartbeats_until_dead:
-                if get_closing_grace_period() is not None:
-                    # already in grace period, no need to set again
-                    continue
-                logger.info(
-                    "%r closing websocket because of heartbeat misses",
-                    session_id,
+            try:
+                await send_transport_message(msg, ws, websocket_closed_callback)
+                commit(msg)
```
jackyzha0, line 81: what is commit doing

src/replit_river/error_schema.py
```diff
@@ -17,9 +17,16 @@
 # ERROR_CODE_CANCEL is the code used when either server or client cancels the stream.
 ERROR_CODE_CANCEL = "CANCEL"
 
+# ERROR_CODE_SESSION_CLOSED is the code used when either server or client closes
+# the session.
+ERROR_CODE_SESSION_CLOSED = "CLOSED"
```
jackyzha0, line 22: this is not in the protocol, UNEXPECTED_DISCONNECT is probably the closest semantically?

src/replit_river/common_session.py
```diff
@@ -29,107 +33,69 @@ async def __call__(
...
+    websocket_closed_callback: Callable[[], Coroutine[Any, Any, None]],
+    get_next_pending: Callable[[], TransportMessage | None],
+    commit: Callable[[TransportMessage], None],
 ) -> None:
-    logger.debug("Start heartbeat")
     while True:
-        await asyncio.sleep(heartbeat_ms / 1000)
-        state = get_state()
-        if state != SessionState.ACTIVE:
+        await message_enqueued.acquire()
```
jackyzha0, line 64: why not use context manager style for automatic release here?

src/replit_river/server_session.py
```diff
@@ -22,18 +21,18 @@
...
     ACK_BIT,
-    STREAM_CLOSED_BIT,
     STREAM_OPEN_BIT,
     GenericRpcHandlerBuilder,
     TransportMessage,
     TransportMessageTracingSetter,
 )
 
-logger = logging.getLogger(__name__)
+STREAM_CLOSED_BIT = 0x0004  # Synonymous with the cancel bit in v2
```
jackyzha0, line 30: seems weird that this is defined away from where all the control flag bits are defined

src/replit_river/server_transport.py
```diff
@@ -33,6 +32,8 @@
 from replit_river.session import Session
 from replit_river.transport_options import TransportOptions
 
+PROTOCOL_VERSION = "v1.1"
```
jackyzha0, line 35: maybe a comment here that only client is v2, server is still v1.1

src/replit_river/session.py
```diff
@@ -96,7 +124,11 @@ def increment_and_get_heartbeat_misses() -> int:
                 self.session_id,
                 self._transport_options.heartbeat_ms,
                 self._transport_options.heartbeats_until_dead,
-                lambda: self._state,
+                lambda: (
+                    self._state
+                    if self._ws_wrapper.ws_state == WsState.OPEN
+                    else SessionState.CONNECTING
```
jackyzha0, line 130: a session wont always have a ws right? what about the SessionState.PENDING state?

src/replit_river/session.py
```diff
@@ -306,3 +338,83 @@ async def close(self) -> None:
...
+            )
+            # session is closing / closed, no need to send heartbeat anymore
+            return
+        try:
+            await send_message(
+                stream_id="heartbeat",
+                # TODO: make this a message class
+                # https://github.com/replit/river/blob/741b1ea6d7600937ad53564e9cf8cd27a92ec36a/transport/message.ts#L42
+                payload={
+                    "ack": 0,
```
jackyzha0, line 400:
```suggestion
                    "type": "ACK",
```

src/replit_river/codegen/client.py
```diff
@@ -118,7 +119,7 @@ class RiverNotType(BaseModel):
 
 class RiverProcedure(BaseModel):
     init: RiverType | None = Field(default=None)
```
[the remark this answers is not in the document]
blast-hardcheese: This is to support both v1 and v2 simultaneously, we do the asserts down below

src/replit_river/codegen/client.py
```diff
@@ -798,10 +806,314 @@ def __init__(self, client: river.Client[{handshake_type}]):
...
+          {binding}: {render_type_expr(tpe)},
+          timeout: datetime.timedelta,
+        ) -> {render_type_expr(output_type)}:
+          return await self.client.send_rpc(
+            {repr(schema_name)},
+            {repr(name)},
+            {binding},
+            {reindent("                    ", render_method)},
+            {reindent("                    ", parse_output_method)},
+            {reindent("                    ", parse_error_method)},
```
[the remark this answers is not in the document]
blast-hardcheese: lmao we format, but due to semantic whitespace we have to ensure that we're _at least_ at the current level of whitespace.

```python
def foo():
    if True:
                  print("This is syntactically valid")
    else:
  print("This is an error")
```

Hold your applause please.

src/replit_river/codegen/client.py
blast-hardcheese: In the current world where we throw exceptions all over the place, this muddies the water. I'd like to consolidate, but post v2.

src/replit_river/codegen/run.py
blast-hardcheese: in fact it does

src/replit_river/common_session.py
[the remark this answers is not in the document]
blast-hardcheese: Pending means we're not connecting yet. It's equivalent in intent to `NO_CONNECTION` so I'll rename. Thanks!

src/replit_river/common_session.py
blast-hardcheese: 🙈 I thought I'd already fixed this, thanks for catching

blast-hardcheese: oh, I did! This is dead code! Deleting

src/replit_river/common_session.py
blast-hardcheese: Dead code! Deleting!

src/replit_river/common_session.py
blast-hardcheese: Dead code! Deleting!

In the other version of this function, `commit` "accepts" a message as having been sent.

src/replit_river/error_schema.py
blast-hardcheese: We're not sending this over the wire, we are forced to give a reason to the python code calling this method. `UNEXPECTED_DISCONNECT` doesn't make sense in the case where we intentionally closed the session but a consumer still tried to call a method post-close

jackyzha0: im fine leaving it as is, maybe just make the doc comment more clear this is a synthetic message that isnt from the server

src/replit_river/server_session.py
```diff
@@ -173,18 +192,23 @@ async def _handle_messages_from_ws(self, tg: asyncio.TaskGroup) -> None:
...
+        tg: asyncio.TaskGroup,
+    ) -> Channel | IgnoreMessage:
         if not msg.serviceName or not msg.procedureName:
-            raise IgnoreMessageException(
-                f"Service name or procedure name is missing in the message {msg}"
+            logger.warning(
+                "Service name or procedure name is missing in the message %r",
+                msg,
             )
+            return IgnoreMessage()
```
masad-frost, line 202: this should return invalid in v2

masad-frost: oh nvm lol

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1280 @@
...
+    # message state
+    _process_messages: asyncio.Event
+    _space_available: asyncio.Event
+
+    # stream for tasks
+    _streams: dict[str, Channel[Any]]
+
+    # book keeping
+    _ack_buffer: deque[TransportMessage]
+    _send_buffer: deque[TransportMessage]
```
masad-frost, line 126: I assume we're using a deque here because .pop(0) reshifts the underlying array?

src/replit_river/server_session.py
masad-frost: It's because it has a different meaning in v1 and v2

masad-frost (commented):
The v1 changes are throwing me off :cry:

Are those bug fixes that need to go into another PR?

src/replit_river/error_schema.py
masad-frost: Yeah also caught my eye. Also fine either way, we haven't had a situation where we needed to differentiate in the TS side because things are pretty automatic as far as re-sending goes. We had to differentiate in crosis because we didn't have the right API to make retries automatic-as-needed.

masad-frost: by automatic i mean it's transaction-like, because the consumer is typically react, things rerun in a transaction-like fashion.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1280 @@
...
+                # already in grace period, no need to set again
+                continue
+            logger.info(
+                "%r closing websocket because of heartbeat misses",
+                session_id,
+            )
+            await close_websocket()
+
+
+async def _serve(
```
masad-frost, line 1111: just out of curiosity, why not a method instead of passing things in?

masad-frost (commented):
is the code for transparent reconnect missing? I can't find where that's happening :grimacing:

src/replit_river/v2/session.py
blast-hardcheese: dequeue because append and popleft are both O(1), `list` is O(n) for shifting the whole buffer.

List would be fast enough, but it's still wasteful when there's a stdlib collection that does what we want.

src/replit_river/v2/session.py
blast-hardcheese: easier to reason about individual functions that have discrete interfaces and conform to their exposed interface. The functions that do the work should further be composed of known quantities.

masad-frost (commented):
i have mixed feelings about the approach in session (but that could be my javascript-rotted brain talking).

it feels like maintaining a dedicated send-loop plus explicit events and queues adds extra complexity, more state, more coordination, especially in the happy path, and maybe even just as much bookkeeping for reconnects/unhappy paths. i wonder if we could just simplify things by having a fire-and-forget-ish approach (like just spawning a task per send and awaiting ws.send directly), only keeping around unacked messages for transparent reconnects when that's implemented. We'd need to block while we're connecting, but we're doing that anyway.

the main thing we'd lose is explicit backpressure, but i'm guessing the underlying websocket implementation already awaits and has backpressure all the way down to the tcp buffer anyway. and honestly, i doubt consumers will really take advantage of explicit backpressure. having an implicit unbounded buffer (which sort of already exists via the event loop) might not even be bad here

src/replit_river/common_session.py
blast-hardcheese: If we never reconnect, `close()` now sends an event to wake up these functions so we can get caught by

```python
        if not ws:
            logger.debug("_buffered_message_sender: ws is not connected, loop")
            continue
```

which `continues` back to the outer loop, which terminates due to

```python
    while our_task and not our_task.cancelling() and not our_task.cancelled():
```

... being false.

If we have a connection interruption, the outer loop will not be false and we'll go back into `await block_until_connected()`.

src/replit_river/common_session.py
blast-hardcheese: if `get_next_pending()` is available, we `await send_transport_message(...)`. That must complete before we move on to the next message. Once we've exhausted the buffer, `await block_until_message_available()` will block until the next message is sent.

blast-hardcheese:
The main advantage I see with the current queue-based approach is consolidating the pipeline for better reasoning about transparent reconnects.

> is the code for transparent reconnect missing? I can't find where that's happening 😬

Both the heartbeat's `close_websocket`...

https://github.com/replit/river-python/blob/db33253465446e4111489cf928ff0f6ca98cc191/src/replit_river/v2/session.py#L487

... and `serve_responses`'s `connection_interrupted` ...
https://github.com/replit/river-python/blob/db33253465446e4111489cf928ff0f6ca98cc191/src/replit_river/v2/session.py#L524

both call out to `client_transport`'s `_retry_connection`
https://github.com/replit/river-python/blob/db33253465446e4111489cf928ff0f6ca98cc191/src/replit_river/v2/client_transport.py#L72-L76

which calls back to `ensure_connected`
https://github.com/replit/river-python/blob/db33253465446e4111489cf928ff0f6ca98cc191/src/replit_river/v2/client_transport.py#L65-L70

blast-hardcheese:
> The v1 changes are throwing me off 😢
> 
> Are those bug fixes that need to go into another PR?

I can split those out. river-babel doesn't pass on Python anymore without bug fixes, and I didn't want to deal with merge conflicts

src/replit_river/session.py
blast-hardcheese: this is v1

src/replit_river/session.py
blast-hardcheese: this is v1 as well. To Faris' point, it's hard to remember that the only v2 changes that matter are in the `v2` package

src/replit_river/server_transport.py
blast-hardcheese: indicated by the package and the presence of `PROTOCOL_VERSION` in the different files where it matters, yes.

src/replit_river/error_schema.py
blast-hardcheese: I think the primary footgun is that there's no distinct error channel for these synthetic errors, so I overcommunicated here. If there end up being more places where we need more synthetic errors we can create a distinct channel for them, distinct from the `RiverServiceException` hierarchy.

src/replit_river/v2/session.py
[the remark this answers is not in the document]
blast-hardcheese: Resolved

src/replit_river/v2/session.py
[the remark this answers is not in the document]
blast-hardcheese: Resolved.

Errors from the `buffered_message_sender` are now communicated back to the calling thread, which now emit `CANCEL` events (with the correct schema).

tests/v2/fixtures.py
```diff
@@ -0,0 +1,219 @@
...
+    async with serve(handle, "localhost") as server:
+        for sock in server.sockets:
+            if (pair := sock.getsockname())[0] == "127.0.0.1":
+                ipv4_laddr = "ws://%s:%d" % pair
+        serve_forever = asyncio.create_task(server.serve_forever())
+        assert ipv4_laddr
+        yield (
+            server_to_client,
+            client_to_server,
+            UriAndMetadata(uri=ipv4_laddr, metadata=None),
```
blast-hardcheese, line 72: ipv4 technically doesn't matter, I just wanted to prevent ping-ponging between v4 and v6 depending on which socket sorts first.

tests/v2/fixtures.py
```diff
@@ -0,0 +1,219 @@
...
+
+@pytest.fixture
+async def bound_client(
+    raw_websocket_meta: tuple[
+        ClientToServerChannel,
+        ServerToClientChannel,
+        UriAndMetadata[None],
+    ],
+    expected: deque[TestTransport],
+) -> AsyncGenerator[Client[None], None]:
```
blast-hardcheese, line 86: This is a fly-by-wire'd websocket and `river.v2.Client[None]` that operates entirely based on the passed-in `expected` test script.

There are still some changes that I'd like to implement in the future, namely when you look at the actually written tests you'll see that `ack=0` for all the `FromClient` messages, it's just because the client sends as fast as it can and it gets ahead of the server acks.

To offer some guarantees we just ensure that the `to_client` and `from_client` queues are fully drained as part of the test success criteria.

tests/v2/fixtures.py
```diff
@@ -0,0 +1,219 @@
...
+    #
+    # watchdog_task = asyncio.create_task(watchdog())
+
+    async def uri_and_metadata_factory() -> UriAndMetadata[None]:
+        return uri_and_metadata
+
+    client = Client(
+        uri_and_metadata_factory=uri_and_metadata_factory,
+        client_id="client-001",
+        server_id="server-001",
```
blast-hardcheese, line 175: Probably not super great to be hard-coded. Future versions of this test harness may expose some kind of Client Builder so that we can do request/response interleaving, should that be necessary to express.

tests/v2/test_v2_rpc.py
```diff
@@ -0,0 +1,97 @@
...
+    if not _AlreadyGenerated:
+        validate_codegen(
+            snapshot=snapshot,
+            snapshot_dir="tests/v2/codegen/snapshot/snapshots",
+            read_schema=lambda: open("tests/v2/test_rpc.schema.json"),
+            target_path="test_basic_rpc",
+            client_name="StreamClient",
+            protocol_version="v2.0",
+        )
+        _AlreadyGenerated = True
```
blast-hardcheese, line 42: This flag is _probably_ not necessary. Tests are executed sequentially, so even though we'd be overwriting the client with itself, there would never be in-process writes.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+            try:
+                result = await output.get()
+            except ChannelClosed as e:
+                raise RiverServiceException(
+                    ERROR_CODE_STREAM_CLOSED,
+                    "Stream closed before response",
+                    service_name,
+                    procedure_name,
+                ) from e
+            except Exception as e:
```
jackyzha0, line 715: when does this exception case happen?

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+                WebsocketClosedException,
+                FailedSendingMessageException,
+            ) as e:
+                raise RiverException(
+                    ERROR_HANDSHAKE,
+                    "Handshake failed, conn closed while sending response",
+                ) from e
+
+            startup_grace_deadline_ms = await get_current_time() + 60_000
+            while True:
```
jackyzha0, line 972: there should only be one message as a response, why loop?

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+                    raise RiverException(
+                        ERROR_HANDSHAKE,
+                        "Handshake failed, conn closed while waiting for response",
+                    ) from e
+
+                try:
+                    response_msg = parse_transport_msg(data)
+                    if isinstance(response_msg, str):
+                        logger.debug(
+                            "_do_ensure_connected: Ignoring transport message",
```
jackyzha0, line 995: yeah this should fatally close

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+        do_close()
+        raise RiverException(
+            ERROR_HANDSHAKE,
+            f"Failed to create ws after retrying {attempt_count} number of times",
+        ) from last_error
+
+    return None
+
+
+async def _recv_from_ws(
```
jackyzha0, line 1057: seems weird that recv_from_ws is responsible for reconnecting

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+                connection_attempts = 0
+                # decode=False: Avoiding an unnecessary round-trip through str
+                # Ideally this should be type-ascripted to : bytes, but there
+                # is no @overrides in `websockets` to hint this.
+                try:
+                    message = await ws.recv(decode=False)
+                except ConnectionClosed:
+                    # This triggers a break in the inner loop so we can get back to
+                    # the outer loop.
+                    await transition_no_connection()
```
jackyzha0, line 1113: could this cause a transition from terminal -> no connection?

src/replit_river/message_buffer.py
```diff
@@ -50,13 +52,22 @@ def peek(self) -> TransportMessage | None:
...
+    async def block_until_message_available(self) -> None:
+        """Allow consumers to avoid spinning unnecessarily"""
+        await self._has_messages.wait()
+
     async def close(self) -> None:
         """
         Closes the message buffer and rejects any pending put operations.
         """
         self._closed = True
+        self._has_messages.set()
```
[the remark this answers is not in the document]
blast-hardcheese: to wake up the `buffered_message_sender` so it can close gracefully, I'll add a comment.

src/replit_river/v2/client.py
```diff
@@ -0,0 +1,205 @@
...
+        f"river.client.{procedure_type}.{service_name}.{procedure_name}",
+        kind=SpanKind.CLIENT,
+    )
+    span_handle = _SpanHandle(span)
+    try:
+        yield span_handle
+    except GeneratorExit:
+        # This error indicates the caller is done with the async generator
+        # but messages are still left. This is okay, we do not consider it an error.
+        raise
```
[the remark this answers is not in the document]
blast-hardcheese: re-raises the caught exception without changing its originating line

src/replit_river/v2/session.py
blast-hardcheese: it's an [artifact of the `websockets` library](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#open-and-closed), for better or worse. Both `send()` and `recv()` throw `ConnectionClosed`, so it really should be handled in both cases.

src/replit_river/common_session.py
```diff
@@ -17,24 +27,80 @@ async def __call__(
...
+            except WebsocketClosedException as e:
+                logger.debug(
+                    "_buffered_message_sender: Connection closed while sending "
+                    "message %r, waiting for retry from buffer",
+                    type(e),
+                    exc_info=e,
+                )
+            except FailedSendingMessageException:
+                logger.error(
+                    "Failed sending message, waiting for retry from buffer",
```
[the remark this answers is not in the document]
blast-hardcheese: The `buffered_message_sender` is tied to the session; it just waits until we have a connection again to wake up and send the buffered messages. This lets us enqueue new messages while we're not connected and we don't run into that out-of-order issue.

Additionally, if either the handshake message doesn't pass muster (seq or ack don't match) then the `close_session` callback is called and the whole session and buffer is destroyed.

src/replit_river/codegen/client.py
```diff
@@ -798,11 +806,291 @@ def __init__(self, client: river.Client[{handshake_type}]):
...
+          )
+                    """,
+                    )
+                ]
+            )
+        else:
+            assert_never(protocol_version)
+
+    current_chunks.append("")
```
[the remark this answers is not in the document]
blast-hardcheese: It gets translated into a newline to avoid sibling classes all getting run together

src/replit_river/v2/client.py
```diff
@@ -0,0 +1,205 @@
...
+    except RiverException as e:
+        span.record_exception(e, escaped=True)
+        _record_river_error(span_handle, RiverError(code=e.code, message=e.message))
+        raise e
+    except BaseException as e:
+        span.record_exception(e, escaped=True)
+        span_handle.set_status(StatusCode.ERROR, f"{type(e).__name__}: {e}")
+        raise e
+    finally:
```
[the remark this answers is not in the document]
blast-hardcheese: oh god I just copied that from the existing client. will fix.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+    payload: Any
+
+
+class ErrorPayload(TypedDict):
+    code: str
+    message: str
+
+
+class ResultError(TypedDict):
+    # Account for structurally incoherent payloads
```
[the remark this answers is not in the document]
blast-hardcheese: The comment refers to the `NotRequired` in `ok: NotRequired[Literal[False]]`, which lets us treat anything that doesn't explicitly have `["ok"] == True` as an error

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+                enqueue_message=self._enqueue_message,
+            )
+        )
+
+    @asynccontextmanager
+    async def _with_stream(
+        self,
+        session_id: str,
+        maxsize: int,
+    ) -> AsyncIterator[tuple[Channel[Exception | None], Channel[ResultType]]]:
```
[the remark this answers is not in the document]
blast-hardcheese: split the error channel out entirely:

```python
    ) -> AsyncIterator[tuple[_BackpressuredWaiter, AsyncIterator[ResultType]]]:
```

now callers can `await backpressured_waiter()` or `async for thing in output:` and both cases will correctly raise out-of-bounds exceptions without having to thread boilerplate everywhere (optionally forgetting to check)

src/replit_river/v2/session.py
blast-hardcheese: Really seems to only be if the `Session` is closed. I'll remove since the underlying exception should be sufficient.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+            except (
+                WebsocketClosedException,
+                FailedSendingMessageException,
+            ) as e:
+                raise RiverException(
+                    ERROR_HANDSHAKE,
+                    "Handshake failed, conn closed while sending response",
+                ) from e
+
+            startup_grace_deadline_ms = await get_current_time() + 60_000
```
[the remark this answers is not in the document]
blast-hardcheese: 👍 Used the existing `handshake_timeout_ms`

src/replit_river/v2/session.py
blast-hardcheese: Is this not necessary, or is receiving a transport message here an error?
```
                    if isinstance(response_msg, str):
                        logger.debug(
                            "_do_ensure_connected: Ignoring transport message",
                            exc_info=True,
                        )
                        continue
```

blast-hardcheese: Oh, just saw your follow-up question

src/replit_river/v2/session.py
blast-hardcheese: Done!

src/replit_river/v2/session.py
blast-hardcheese: All of the transition functions are guarded:

```python
        async def transition_no_connection() -> None:
            if self._state in TerminalStates:
                return
```

src/replit_river/codegen/client.py
```diff
@@ -798,11 +806,291 @@ def __init__(self, client: river.Client[{handshake_type}]):
...
+            {reindent("                    ", parse_output_method)},
+            {reindent("                    ", parse_error_method)},
+          )
+                    """,
+                    )
+                ]
+            )
+        elif protocol_version == "v2.0":
+            raise ValueError(
+                "It is expected that protocol v2 uploads have both init and input "
```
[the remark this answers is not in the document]
blast-hardcheese: It's just part of the codegen, it's more important that they communicate something to the user. They never end up in datadog.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1239 @@
...
+        for error_channel, stream in self._streams.values():
+            stream.close()
+            # Wake up backpressured writers
+            await error_channel.put(
+                SessionClosedRiverServiceException(
+                    "river session is closed",
+                )
+            )
+        # Before we GC the streams, let's wait for all tasks to be closed gracefully.
+        await asyncio.gather(*[stream.join() for _, stream in self._streams.values()])
```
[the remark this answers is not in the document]
blast-hardcheese: I just added a timeout that logs the `span_ids` in case of a deadlock. It _shouldn't_, but if it does we'll be able to figure out why.

src/replit_river/v2/session.py
```diff
@@ -0,0 +1,1248 @@
...
+        """Sends a subscription request to the server.
+
+        Expects the input and output be messages that will be msgpacked.
+        """
+        stream_id = nanoid.generate()
+        await self._enqueue_message(
+            service_name=service_name,
+            procedure_name=procedure_name,
+            stream_id=stream_id,
+            control_flags=STREAM_OPEN_BIT,
```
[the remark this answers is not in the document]
blast-hardcheese: Found it in the [PROTOCOL.md](https://github.com/replit/river/blob/86ce70abac5397636bfc80a7085a7f316aa63abf/PROTOCOL.md#subscription) -- I think the semantics there are wrong, we should `OPEN|CLOSE` initially, and rely on the remote to send. Relying on the client to ack close is unnecessarily awkward.

blast-hardcheese:
Five remaining flakes in babel:

```
flaked:
- StreamDisconnectNotifs
- RpcDisconnectNotifs
- ShouldNotSendBufferAfterSessionDisconnect
- ComplexSituation
- UploadDisconnectNotifs
```

PR #107, opened 2024-11-07
[chore] Switch to uv

blast-hardcheese:
Why
===

Consistency with other tooling

What changed
============

- Switched to uv from poetry

Test plan
=========

Does CI work? Does releasing publish a new version with the correct contents?


.github/workflows/python-publish.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install --upgrade poetry
+        pip install --upgrade uv
```
ryantm, line 31: Can we also use the uv action here?

ryantm (approved):

.github/workflows/ci.yml
```diff
@@ -7,38 +7,33 @@ jobs:
...
       - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          enable-cache: true
+
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
```
ryantm, line 17: Can we let uv manage Python too?

.github/workflows/python-publish.yml
blast-hardcheese: Sure can!

.github/workflows/ci.yml
blast-hardcheese: I'm inclined to use `setup-python`, as is recommended by https://docs.astral.sh/uv/guides/integration/github/#setting-up-python, since GitHub manages caching

blast-hardcheese: If you feel strongly about this we can change it afterwards, but let's see if publishing works

PR #111, opened 2024-11-19
feat/maybe async handlers

blast-hardcheese:
Why
===

There are a bunch of needless LSP errors because we do `rpc_method_handler(servicer.Foo, ...)`, where `Foo` returns `ResponseType | Awaitable[ResponseType]`. These types are a little convoluted, but it makes it possible to use grpc-generated "servicer" as-is.

What changed
============

Add synchronous response types for river server handlers

Test plan
=========

I manually patched these methods in production codebases and everything typechecks.

### DIFF
--- a/replit_river/rpc.py
+++ b/replit_river/rpc.py
@@ -193,7 +193,9 @@ def get_response_or_error_payload(
 
 
 def rpc_method_handler(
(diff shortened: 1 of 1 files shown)


ryantm (approved):
The changes look okay for the stated purpose.

I'm a little concerned this will result in more people accidentally using synchronous code when they should have been using async or mixing the two by mistake. Is there an alternative where we lint this and fix the call sites to all be async?

PR #113, opened 2024-11-20
chore/decompose monolithic client gen

blast-hardcheese:
Why
===

Some clients end up with tens of thousands of lines per file. This causes some difficulty for typecheckers (in particular, pyright). Decompose into services to get some breathing room.

What changed
============

- Client codegen decomposition

Test plan
=========

Does typechecking improve? ✅


masad-frost (approved):
🙈

PR #110, opened 2024-11-15
Implement interceptors to support OTel tracing

cbrewster:
Why
===

We'd like to see river spans, this adds support for a basic interceptor framework, this let's us build layers for tracings, metrics, and logging.

What changed
============

- Add `ClientInterceptor` abstract base class to define the interceptor interface
    - Each interceptor calls the next interceptor (or the actual underlying procedure). This gives it a change to look at both inputs and outputs.
- Added `OpenTelemetryClientInterceptor` which provides spans for all procedure calls

Test plan
=========

- Wire up the otel interceptor to your application
- You should see pid2 spans show up!

PR #115, opened 2024-11-21
Add opentelemetry instrumentation

cbrewster:
Why
===

Supercedes #110, we want to punt desingin interceptors until later.

We'd like to have tracing between client and server, this implements spans for the client side.

What changed
============

- Add tracing spans for each procedure call
- Add support for propagating the tracing context to the river server using the tracing field of the transport message
    - Ideally the trasnsport message wouldn't hardcode the tracing fields, but we'd need to fix that in the JS implementation

Test plan
=========

- Connect this up to our internal usage, we should see spans for river procedure calls

### DIFF
--- a/replit_river/rpc.py
+++ b/replit_river/rpc.py
@@ -22,6 +22,7 @@ from typing import (
 
 import grpc
 from aiochannel import Channel, ChannelClosed
+from opentelemetry.propagators.textmap import Setter
 from pydantic import BaseModel, ConfigDict, Field
 
 from replit_river.error_schema import (
@@ -86,6 +87,11 @@ class ControlMessageHandshakeResponse(BaseModel):
     status: HandShakeStatus
 
 
+class PropagationContext(BaseModel):
+    traceparent: Optional[str] = None
+    tracestate: Optional[str] = None
+
+
 class TransportMessage(BaseModel):
     id: str
     # from_ is used instead of from because from is a reserved keyword in Python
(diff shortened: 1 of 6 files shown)


cbrewster:

* main:
  * **PR #115** 👈

scripts/lint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
blast-hardcheese, line 1: ope, good catch

replit_river/rpc.py
```diff
@@ -86,6 +87,11 @@ class ControlMessageHandshakeResponse(BaseModel):
     status: HandShakeStatus
 
 
+class PropagationContext(BaseModel):
+    traceparent: str
+    tracestate: str
```
blast-hardcheese, line 92: Based on below, will this mean that one of these two values will always be `""`? Would it be preferable to have them be `NotRequired[str]` instead?

blast-hardcheese (approved):
One question, other than that 📈

replit_river/rpc.py
cbrewster: I'll go ahead and give them some default values. Both _should_ be set, but the way otel injects these at runtime is it sets each field separately so there will be a point in time where one is set and the other isn't.

tbh I think the protocol should just have an arbitrary `metadata` field on the transport message that is `dict[str, str]`, that decouples the river protocol from specific tracing injection formats. But that's for another day...

cbrewster: Err last minute decided to just make these optional like all of our other similar fields, rather than defaulting to empty str

PR #117, opened 2024-11-21
bugfix: Ensure both trace fields are set if tracing is used

cbrewster:
Why
===

If `tracing` is set, we expect `traceparent` and `tracestate` to be non-null on the server side. If there is no tracestate, an empty string is sufficient.

What changed
============

- Default `traceparent` and `tracestate` to empty strings

Test plan
=========

- Tracing should work without the TS server rejecting these messages


cbrewster:

* main:
  * **PR #117** 👈

blast-hardcheese (approved):
ope

PR #120, opened 2024-11-22
bugfix: make streaming spans last for the entire duration of the stream

cbrewster:
Why
===

The streaming spans were immediately finishing because we were only tracing the construction of the `AsyncIterator` and not the full iteration.

What changed
============

- loop and yield each message of the streaming procedures so the span doesn't finish until the async iterator is disposed.

Test plan
=========

- Streaming procedure spans should now last the full duration of the stream procedure


cbrewster:

* main:
  * **PR #120** 👈

blast-hardcheese (approved):
Do'h, good catch

PR #121, opened 2024-11-23
bugfix: include streaming procedure errors in traces

cbrewster:
Why
===

Streaming response procedures don't raise errors like the other procedures, instead the errors are included in the `AsyncIterator`.

Also fixes issues with using contextvars + async generators.

What changed
============

- Check if the message from the async iterator is a `RiverError`, if so, record it on the span
- Use `start_span` instead of `start_as_current_span`, the latter resets a contextvar in its `finally` clause which is invalid to do for async generators as the async generator's finalizers run in a different context
    - Thread the span through manually so we still propagate the tracing info

Test plan
=========

- Should see errors for failed streaming procedures
- Logs about resetting contextvars in a different context should go away
- Added some tests for the otel stuff to make sure the error handling works here

### DIFF
--- a/replit_river/rpc.py
+++ b/replit_river/rpc.py
@@ -388,8 +388,9 @@ def upload_method_handler(
 
             convert_inputs_task = task_manager.create_task(_convert_inputs())
             convert_outputs_task = task_manager.create_task(_convert_outputs())
-            await asyncio.wait((convert_inputs_task, convert_outputs_task))
-
+            done, _ = await asyncio.wait((convert_inputs_task, convert_outputs_task))
+            for task in done:
+                await task
         except Exception as e:
             logger.exception("Uncaught exception in upload")
             await output.put(
@@ -440,17 +441,16 @@ def stream_method_handler(
             response = method(request, context)
 
             async def _convert_outputs() -> None:
-                try:
-                    async for item in response:
-                        await output.put(
-                            get_response_or_error_payload(item, response_serializer)
-                        )
-                finally:
-                    output.close()
+                async for item in response:
+                    await output.put(
+                        get_response_or_error_payload(item, response_serializer)
+                    )
 
             convert_inputs_task = task_manager.create_task(_convert_inputs())
             convert_outputs_task = task_manager.create_task(_convert_outputs())
-            await asyncio.wait((convert_inputs_task, convert_outputs_task))
+            done, _ = await asyncio.wait((convert_inputs_task, convert_outputs_task))
+            for task in done:
+                await task
         except grpc.RpcError:
             logger.exception("RPC exception in stream")
             code = grpc.StatusCode(context._abort_code).name if context else "UNKNOWN"
--- a/replit_river/session.py
+++ b/replit_river/session.py
@@ -38,6 +38,9 @@ from .rpc import (
 
 logger = logging.getLogger(__name__)
 
+trace_propagator = TraceContextTextMapPropagator()
+trace_setter = TransportMessageTracingSetter()
+
 
 class SessionState(enum.Enum):
     """The state a session can be in.
@@ -386,9 +389,7 @@ class Session(object):
         )
         if span:
             with use_span(span):
-                TraceContextTextMapPropagator().inject(
-                    msg, None, TransportMessageTracingSetter()
-                )
+                trace_propagator.inject(msg, None, trace_setter)
         try:
             # We need this lock to ensure the buffer order and message sending order
             # are the same.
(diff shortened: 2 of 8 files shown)


cbrewster:

* main:
  * **PR #121** 👈

masad-frost:
@blast-hardcheese can I dump this on you please? :D

replit_river/session.py
```diff
@@ -382,9 +384,11 @@ async def send_message(
             serviceName=service_name,
             procedureName=procedure_name,
         )
-        TraceContextTextMapPropagator().inject(
-            msg, None, TransportMessageTracingSetter()
-        )
+        if span:
+            with use_span(span):
```
cbrewster, line 388: Unfortunately the Propagator requires the span to be in the context and you can't just pass it directly. Should be fine because there's no await point inside the `with`

replit_river/rpc.py
```diff
@@ -440,17 +441,16 @@ async def _convert_inputs() -> None:
             response = method(request, context)
 
             async def _convert_outputs() -> None:
-                try:
-                    async for item in response:
-                        await output.put(
-                            get_response_or_error_payload(item, response_serializer)
-                        )
-                finally:
-                    output.close()
```
cbrewster, line 449: Output is closed on L486, so I removed this close.

replit_river/rpc.py
```diff
@@ -440,17 +441,16 @@ async def _convert_inputs() -> None:
...
+                    await output.put(
+                        get_response_or_error_payload(item, response_serializer)
+                    )
 
             convert_inputs_task = task_manager.create_task(_convert_inputs())
             convert_outputs_task = task_manager.create_task(_convert_outputs())
-            await asyncio.wait((convert_inputs_task, convert_outputs_task))
+            done, _ = await asyncio.wait((convert_inputs_task, convert_outputs_task))
+            for task in done:
+                await task
```
cbrewster, line 453: `asyncio.wait` doesn't raise exceptions from the inner tasks, you have to await them individually for the exceptions to be raised. This is definitely a bug in our current river server implementation.

blast-hardcheese: 😱

replit_river/session.py
```diff
@@ -382,9 +384,11 @@ async def send_message(
...
+                TraceContextTextMapPropagator().inject(
```
blast-hardcheese, line 389: would it be worth it (or is it advantageous) to cache these propagators/setters?

tests/river_fixtures/logging.py
```diff
@@ -15,8 +15,15 @@ class NoErrors:
 
     def __init__(self, caplog: LogCaptureFixture):
         self.caplog = caplog
+        self._allow_errors = False
+
+    def allow_errors(self) -> None:
+        self._allow_errors = True
```
blast-hardcheese, line 21: if only it were always this easy

tests/test_opentelemetry.py
```diff
@@ -0,0 +1,143 @@
...
+    async for response in client.send_subscription(
+        "test_service",
+        "subscription_method",
+        "Bob",
+        serialize_request,
+        deserialize_response,
+        deserialize_error,
+    ):
+        assert isinstance(response, str)
+        assert "Subscription message" in response
```
blast-hardcheese, line 69: Far from me to ask for _more_ convoluted tests, but am I correct in presuming that it's not possible to test exception propagation within the context of OTEL here?

blast-hardcheese (commented):
Trending in the right direction, for sure. Great call-out wrt the `asyncio.wait` not raising the captured exceptions. In a sense, it can't do that sensibly.

Otoh, this really makes me want a linting rule for "you called a function that returned a value, but you didn't receive it." Kind of challenging in Python due to the dynamic and multipurpose nature of many functions, but it would have caught this.

blast-hardcheese (approved):
I intended to approve. Pardon.

replit_river/session.py
cbrewster: Yeah we can probably just save these as globals, they don't store any state or anything, just implement an interface.

tests/test_opentelemetry.py
cbrewster: Say more? The last test in this file does test for errors coming from a stream rpc, I can add erorr tests for the other rpcs though!

blast-hardcheese: I completely missed that that test was testing what I was looking for. I expected it to be of the form

```
    async def stream_data() -> AsyncGenerator[str, None]:
        raise RiverError(...)
        yield "nope"
```

The circuitous path back towards a raised error threw me off. If that's what's required, I'm happy to adjust my expectations and bikeshed locally one of these weekends 😉

cbrewster: Ah thats because `stream_data` is the client->server stream, there's also the server->client stream which is the one that has the error and the path we want to test.

blast-hardcheese: Makes sense. I am mollified.

PR #122, opened 2024-11-25
use empty context in restore budget task

cbrewster:
Why
===

By default, new asyncio tasks hold onto a reference to the current context. If the user of river has a lot of data in the context while the restore budget task is created, that context cannot be GC'd until the task exits. This can lead to accidental memory leaks.

As a followup we will replace this rate limiter with one that does not require background tasks.

What changed
============

- Create a fresh context for restore budget task

Test plan
=========

- Shouldn't have any behavior impact here, we don't rely on the context for budget restoration.
- We should see fewer memory leaks internally

### DIFF
--- a/replit_river/rate_limiter.py
+++ b/replit_river/rate_limiter.py
@@ -1,5 +1,6 @@
 import asyncio
 import random
+from contextvars import Context
 from typing import Dict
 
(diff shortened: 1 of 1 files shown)


cbrewster:

* main:
  * **PR #122** 👈

ryantm (approved):

PR #125, opened 2024-11-28
[tests] handler decomposition

blast-hardcheese:
Why
===

Striking a balance between test handler reuse and colocating the handler specifiers with the tests that use them.

What changed
============

- Decomposing monolithic server handler specifiers into directly DI-ing reusable components
- For tests that need bespoke handlers, defining those alongside the methods that use them.

Test plan
=========

CI


blast-hardcheese:
@cbrewster Following up on an earlier PR comment. This brings handler invocation closer to the tests that use them.

cbrewster (approved):
Nice, I like it

PR #129, opened 2024-12-03
Don't record GeneratorExit errors in stream RPC spans

cbrewster:
Why
===

Stream RPCs which are not read to completion would show up with a `GeneratorExit` error in their associated span. There are valid cases where a caller may not need to exhaustively read the stream, so these should not be considered errors.

What changed
============

- Stop using the span as a context manager, we'll handle recording errors ourselves
- Catch RiverExceptions, Exceptions, and CancelledErrors
- Switch from `AsyncIterator` to `AsyncGenerator` for stream return types
    - This allows us to call close on the generator
    - We may also want to do this in the codegen, this will be especially useful for cancellation support (just close the generator)

Test plan
=========

- Added a test which ensured a closed generator creates a span with OK status

### DIFF
--- a/replit_river/client.py
+++ b/replit_river/client.py
@@ -1,11 +1,12 @@
 import logging
-from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
+from collections.abc import AsyncIterable, Awaitable, Callable
 from contextlib import contextmanager
+from dataclasses import dataclass
 from datetime import timedelta
-from typing import Any, Generator, Generic, Literal, Optional, Union
+from typing import Any, AsyncGenerator, Generator, Generic, Literal, Optional, Union
 
 from opentelemetry import trace
-from opentelemetry.trace import Span, SpanKind, StatusCode
+from opentelemetry.trace import Span, SpanKind, Status, StatusCode
 
 from replit_river.client_transport import ClientTransport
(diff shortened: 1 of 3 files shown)


cbrewster:

* main:
  * **PR #129** 👈

replit_river/client.py
```diff
@@ -160,15 +161,27 @@ def _trace_procedure(
...
+        _record_river_error(span, RiverError(code=e.code, message=e.message))
+        set_status = True
+        raise e
+    except (Exception, CancelledError) as e:
+        span.record_exception(e)
+        span.set_status(StatusCode.ERROR, f"{type(e).__name__}: {e}")
+        set_status = True
+        raise e
+    finally:
+        if not set_status:
```
lhchavez, line 182: maybe add a comment that the `except` chain will _explicitly_ not catch any `BaseException`s (except `CancelledError`. this is peak [enclave](https://www.youtube.com/watch?v=PrJSdYCkoWk) territory) because those are kinda sorta reserved for Python to do control flow.

lhchavez (approved):

replit_river/client.py
cbrewster: Adjusted things to specifically catch and raise `GeneratorExit` to make things more clear.

PR #130, opened 2024-12-04
Address incorrect packaging information

blast-hardcheese:
Why
===

Looks like setuptools has some issues with a new pypi restriction: https://github.com/astral-sh/uv/issues/9513. Hatchling does not have these issues.

What changed
============

Switch to hatchling

Test plan
=========

Can we publish?


cbrewster (approved):
:egg:

PR #131, opened 2024-12-10
Fix accidental truncation of RPC timeouts

cbrewster:
Why
===

`asyncio.timeout` takes a `float` in units of seconds. We were truncating to an `int` so for example a `200ms` timeout would be converted to `0s`

What changed
============

- Remove `int()`

Test plan
=========

- Added some tests to make sure timeouts work and that small timeouts are not rounded to 0


cbrewster:

* main:
  * **PR #131** 👈

PR #132, opened 2024-12-12
Ensure WebSocket is always marked as closed after getting ConnectionClosed error

cbrewster:
Why
===

When we received a `ConnectionClosed` error while reading from the WebSocket, we were not marking the `ws_wrapper` as closed. This meant that the reconnection code thought that there was already an open WebSocket connection and did not attempt reconnection. Leaving the session stuck until the grace period elapses.

What changed
============

- Always call `close_websocket` after getting a `ConnectionClosed` error so that the bookkeeping is up to date
    - This unifies some of the codepaths so we perform the same logic no matter where the disconnect error happens
    - Also renamed `_handle_connection_closed` so its more clear
- `close_websocket` also handles triggering a reconnect, if enabled

Test plan
=========

- TBD

### DIFF
--- a/replit_river/session.py
+++ b/replit_river/session.py
@@ -116,10 +116,15 @@ class Session(object):
         async with self._ws_lock:
             return await self._ws_wrapper.is_open()
 
-    async def _begin_close_session_countdown(self) -> None:
-        """Begin the countdown to close session, this should be called when
-        websocket is closed.
+    async def _handle_connection_closed(self) -> None:
         """
+        Handle the WebSocket connection being closing.
+        This will trigger connection retries, if enabled, and starts a
+        session reconnection timer.
+        If the timer expires before reconnection, the session will be closed.
+        """
+        # ensure websocket is closed and initiate connection retries if applicable.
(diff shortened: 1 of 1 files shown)


cbrewster:

* main:
  * **PR #132** 👈

replit_river/session.py
```diff
@@ -145,12 +150,7 @@ async def serve(self) -> None:
                 try:
                     await self._handle_messages_from_ws(tg)
                 except ConnectionClosed:
-                    if self._retry_connection_callback:
-                        self._task_manager.create_task(
-                            self._retry_connection_callback()
-                        )
-
-                    await self._begin_close_session_countdown()
+                    await self._handle_connection_closed()
```
jackyzha0, line 153: seems like we lost the _retry_connection_callback callback here?

cbrewster: That is already handled down inside `close_websocket`: https://github.com/replit/river-python/blob/235406c66b60a8d446ab429d54cbe60a34cc071f/replit_river/session.py#L454-L455

jackyzha0 (approved):

PR #135, opened 2025-02-06
[client codegen] Unknown enumeration values, attempt 2

blast-hardcheese:
Why
===

While technically the previously emitted deserializers were functional, the ergonomics were basically unusable. I fixed the local development flow so I could complete the whole feature end-to-end this time, confirming that it works.

What changed
============

- Wrapping unknown enumerator parameters in a newly exposed `RiverUnknownValue`, emitted via a newly exposed `raise_unknown`

It would be preferred to have had the `Annotated[...]` directly on underlying models instead of leaking the unknown status into the top-level enumerations, but that was going to require more surgery.

The client migration from this encoding to that encoding will be very easy.

Test plan
=========

### DIFF
--- a/src/replit_river/client.py
+++ b/src/replit_river/client.py
@@ -37,7 +37,7 @@ class RiverUnknownValue(BaseModel):
     value: Any
 
 
-def raise_unknown(
+def translate_unknown_value(
     value: Any, handler: Callable[[Any], Any], info: ValidationInfo
 ) -> Any | RiverUnknownValue:
     try:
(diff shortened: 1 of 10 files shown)


src/replit_river/client.py
```diff
@@ -27,6 +31,21 @@
 tracer = trace.get_tracer(__name__)
 
 
+@dataclass(frozen=True)
+class RiverUnknownValue(BaseModel):
+    tag: Literal["RiverUnknownValue"]
+    value: Any
+
+
+def raise_unknown(
```
ryantm, line 40: I was sort of confused by this name, since the implementation doesn't actually raise anything.

ryantm (approved):

src/replit_river/client.py
blast-hardcheese: I can change it, I was going for "raise an exceptional value" but I can see how it's confusing

blast-hardcheese: Going with `translate_unknown_value`

PR #141, opened 2025-03-14
Stricter error types

airportyh:
Why
===

Have the generated code return error types that actually satisfy the `ErrorType`. Previously the union type for error types generated can be values of `RiverUnknownValue` which is not an error type. This introduces a `RiverUnknownError` and uses that in the error unions instead. Also type adapter types are now more strictly typed.

# Testing

When switching the client code over, we should get errors about `RiverUnknownError` not being handled and force you to fix them.


src/replit_river/client.py
```diff
@@ -46,6 +50,15 @@ def translate_unknown_value(
         return RiverUnknownValue(tag="RiverUnknownValue", value=value)
 
 
+def translate_unknown_error(
+    value: Any, handler: Callable[[Any], Any], info: ValidationInfo
+) -> Any | RiverUnknownError:
+    try:
+        return handler(value)
+    except Exception:
+        return RiverUnknownError(code=ERROR_CODE_UNKNOWN, message="Unknown error")
```
blast-hardcheese, line 59: We have some guarantees about error values

```suggestion
    value: Any, handler: Callable[[Any], Any], info: ValidationInfo
) -> Any | RiverUnknownError:
    try:
        return handler(value)
    except Exception:
        if isinstance(value, dict) and "code" in value and "message" in value:
            return RiverUnknownError(
                code=value["code"],
                message=value["message"],
            )
        else:
            return RiverUnknownError(code=ERROR_CODE_UNKNOWN, message="Unknown error")
```

src/replit_river/codegen/typing.py
```diff
@@ -87,8 +87,8 @@ def render_type_expr(value: TypeExpression) -> str:
         case OpenUnionTypeExpr(inner):
             return (
                 "Annotated["
-                f"{render_type_expr(inner)} | RiverUnknownValue,"
-                "WrapValidator(translate_unknown_value)"
+                f"{render_type_expr(inner)} | RiverUnknownError,"
```
blast-hardcheese, line 90: This has to be based on whether we're `render_type_expr`-ing an error or not

tests/codegen/snapshot/snapshots/test_unknown_enum/enumService/needsEnum.py
```diff
@@ -19,26 +19,40 @@
...
 
-NeedsenumInputTypeAdapter: TypeAdapter[Any] = TypeAdapter(NeedsenumInput)
+NeedsenumInputTypeAdapter: TypeAdapter[NeedsenumInput] = TypeAdapter(NeedsenumInput)
 
 NeedsenumOutput = Annotated[
-    Literal["out_first"] | Literal["out_second"] | RiverUnknownValue,
-    WrapValidator(translate_unknown_value),
+    Literal["out_first"] | Literal["out_second"] | RiverUnknownError,
+    WrapValidator(translate_unknown_error),
 ]
```
blast-hardcheese, line 35: see, this is now a `RiverError` instead of `RiverUnknownValue`

src/replit_river/codegen/typing.py
blast-hardcheese: why were these resolved?

tests/codegen/snapshot/test_enum.py
```diff
@@ -40,12 +40,30 @@
           "errors": {
             "anyOf": [
               {
-                "type": "string",
-                "const": "err_first"
+                "type": "object",
```
airportyh, line 43: @blast-hardcheese I found that the river protocol allowed errors to have any type, including string literals. In order to make the test pass I had to change this test fixture. So to make this change would be imposing additional semantics that's not part of the protocol. Do we really want to?

Alternative could be to make `ErrorType` allow for `RiverUnknownValue` as an option.

blast-hardcheese: I added two tests that verify the functionality

blast-hardcheese (approved):

PR #145, opened 2025-03-18
patch reconnect deadlock

jackyzha0:
Why
===

deadlock in the server-side reconnect case:
1. establish handshake acquires session lock: https://github.com/replit/river-python/blob/main/src/replit_river/server_transport.py#L231
2. if there is a session mismatch, we close the old session: https://github.com/replit/river-python/blob/main/src/replit_river/server_transport.py#L272
3. session.close calls _close_session_callback https://github.com/replit/river-python/blob/main/src/replit_river/session.py#L562 which is bound to transport._delete_session
4. _delete_session also tries to acquire the session lock https://github.com/replit/river-python/blob/main/src/replit_river/transport.py#L44

What changed
============

split up critical sections in server transport similar to client side
1. make `_get_session` acquire lock, dont need to lock for creating the session until we call `_set_session`
2. close session also acquires its own lock

Test plan
=========

added a test

Notes
=========

I have a draft of a more in-depth approach on [this branch](https://github.com/replit/river-python/tree/jackyzha0/fix-deadlock) which uses a lock-ownership-transfer based approach that should catch it more generically but ran into ownership problems lol

Seeing as we are planning on migrating chat service to Node anyways so we only have one River server implementation, we hopefully don't have to maintain this surface for much longer 🤞

### DIFF
--- a/src/replit_river/client_transport.py
+++ b/src/replit_river/client_transport.py
@@ -344,7 +344,6 @@ class ClientTransport(Transport, Generic[HandshakeMetadataType]):
                 # If the session status is mismatched, we should close the old session
                 # and let the retry logic to create a new session.
                 await old_session.close()
-                await self._delete_session(old_session)
 
             raise RiverException(
                 ERROR_HANDSHAKE,
--- a/src/replit_river/server_transport.py
+++ b/src/replit_river/server_transport.py
@@ -236,9 +236,7 @@ class ServerTransport(Transport):
         client_next_expected_seq = (
             handshake_request.expectedSessionState.nextExpectedSeq
         )
-        client_next_sent_seq = (
-            handshake_request.expectedSessionState.nextSentSeq or 0
-        )
+        client_next_sent_seq = handshake_request.expectedSessionState.nextSentSeq or 0
         if old_session and old_session.session_id == handshake_request.sessionId:
             # check invariants
             # ordering must be correct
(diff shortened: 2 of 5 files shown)


src/replit_river/server_transport.py
```diff
@@ -115,44 +137,26 @@ async def _get_or_create_session(
...
 
-            if session_to_close:
-                logger.info("Closing stale session %s", session_to_close.session_id)
-                await session_to_close.close()
+        if old_session and new_session != old_session:
+            logger.info("Closing stale session %s", old_session.session_id)
+            await old_session.close()
+
+        async with self._session_lock:
             self._set_session(new_session)
```
lhchavez, line 158: for safety, one property that the old code had that this one doesn't is that there _could_ be two concurrent calls to this function (with the same id?) and then we would close the old session twice and leak one of the new sessions!

to prevent that what we can do is to rename `_set_session` to `_set_session_locked` (and add a docstring comment that it needs the `session_lock` to operate). and have `_set_session_locked` detect this mid-air collision and ... do something with the preexisting one (close it???)

with that out of the way, this relaxing of the locks' critical sections _seems_ to be safe, because the `Session` by itself is safe to create twice (as long as we close it, otherwise we leak tasks!!!)

src/replit_river/client_transport.py
```diff
@@ -344,7 +344,6 @@ async def _establish_handshake(
                 # If the session status is mismatched, we should close the old session
                 # and let the retry logic to create a new session.
                 await old_session.close()
-                await self._delete_session(old_session)
```
lhchavez, line 347: lol dupes

blast-hardcheese:
@jackyzha0 I applied this delta overtop #147, available at `jacky-deadlock-patch`, so we can keep moving

PR #146, opened 2025-03-18
[bug] Avoiding mixing type rendering and unstructured codegen

blast-hardcheese:
Why
===

Introduced in #142, `NotRequired[A] | None` doesn't typecheck, but we didn't have any codegen tests that exhausted that path.

What changed
============

Move the statically rendered `${render(...)} | None` into `${render(UnionTypeExpr([..., NoneTypeExpr()]))}` so the type renderer knows about and can properly unify/deduplicate/render the structured union.

Test plan
=========

TODO: Add a test that makes it so we don't regress here in the future.


jackyzha0:
can we add a quick codegen test for sanity? looks good otherwise

blast-hardcheese:
Yep -- that's part of the TODO

jackyzha0 (approved):

PR #147, opened 2025-03-19
[chore] feat/dethread clientserver

blast-hardcheese:
Why
===

It's challenging to reason about how all the bits fit together right now.

What changed
============

Dethread server and client sessions and transport.

Test plan
=========

CI


src/replit_river/server_session.py
```diff
@@ -0,0 +1,247 @@
...
+                            logger.warning("no stream for %s", msg.streamId)
+                            raise IgnoreMessageException(
+                                "no stream for message, ignoring"
+                            )
+                        await add_msg_to_stream(msg, stream)
+                    else:
+                        # TODO(dstewart) This looks like it opens a new call to handler
+                        #                on ever ws message, instead of demuxing and
+                        #                routing.
+                        _stream = await self._open_stream_and_call_handler(msg, tg)
```
masad-frost, line 151: :eyes:

masad-frost (approved):
LGTM, every comment i was about to leave was already in existing code and we don't wanna deal with that now lol

masad-frost:
I had attempted to do this, and i found it my stashes, then ended up just doing this https://github.com/replit/river-python/pull/85

src/replit_river/server_session.py
blast-hardcheese: Yeah Connor found that as well -- the comment is stale from when I was first getting up to speed on river

PR #153, opened 2025-04-27
[bug] Address proactive reconnect codepaths

blast-hardcheese:
Why
===

We're seeing some connectivity issues using the river v2 client. Scrutinizing codepaths, there were some missed cases

What changed
============

- `_recv_from_ws` should catch all exceptions inside the loop instead of terminating early and never restarting
- Differentiate between `ConnectionClosedOK` and `ConnectionClosed` in exception handlers. If we close on purpose, don't initiate reconnect
- `session.close()` should block all callers until we are down, instead of just the initiating caller.

Test plan
=========

CI/CD

### DIFF
--- a/src/replit_river/v2/session.py
+++ b/src/replit_river/v2/session.py
@@ -1274,6 +1274,9 @@ async def _recv_from_ws(
                 # is no @overrides in `websockets` to hint this.
                 try:
                     message = await ws.recv(decode=False)
+                except ConnectionClosedOK as e:
+                    close_session(e)
+                    continue
                 except ConnectionClosed:
                     # This triggers a break in the inner loop so we can get back to
                     # the outer loop.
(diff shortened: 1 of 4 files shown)


src/replit_river/v2/session.py
```diff
@@ -397,16 +411,31 @@ async def _enqueue_message(
         self._process_messages.set()
 
     async def close(
-        self, reason: Exception | None = None, current_state: SessionState | None = None
+        self,
+        reason: Exception | None = None,
+        current_state: SessionState | None = None,
+        _wait_for_closed: bool = True,
```
cbrewster, line 417: It seems like we will always wait for closed if the `_closing_waiter` isn't set yet, even if this is set to `False`?

cbrewster (approved):

src/replit_river/v2/session.py
blast-hardcheese: The boolean here was admittedly a hack to keep moving, let me see if I can pull it out

jackyzha0 (approved):
commented irl re: assigning self._terminating_task, rest lgtm!

PR #156, opened 2025-05-06
[v2 bug] Disable websocket max_size

blast-hardcheese:
Why
===

When we hit the maximum size of a websocket packet the websocket client proactively disconnects itself. Unfortunately this can happen outside of the observed lifecycle of the owned socket, so when we attempt to reconnect we just resend the last request and see whether or not we get a valid response.

What changed
============

- Disable `max_size` during `websocket.connect(...)`

Test plan
=========

CI

Without the bugfix commit, test logs include:

```
DEBUG    replit_river.common_session:common_session.py:102 _buffered_message_sender: Sent 'jKAcLL2O3XBL4xZF1cEWv'
DEBUG    websockets.server:protocol.py:572 < BINARY 8b a2 69 64 b5 6a 4b 41 63 4c 4c 32 4f 33 58 42 ... 70 61 79 6c 6f 61 64 80 [244 bytes]
DEBUG    websockets.server:protocol.py:719 > BINARY 88 a2 69 64 b5 52 42 59 4b 6a 42 2d 63 59 37 5f ... 61 61 61 61 61 61 61 61 [1048711 bytes]
DEBUG    websockets.client:protocol.py:719 > CLOSE 1009 (message too big) over size limit (? > 1048576 bytes) [37 bytes]
DEBUG    websockets.client:protocol.py:169 = connection is CLOSING
DEBUG    websockets.server:protocol.py:572 < CLOSE 1009 (message too big) over size limit (? > 1048576 bytes) [37 bytes]
DEBUG    websockets.server:protocol.py:719 > CLOSE 1009 (message too big) over size limit (? > 1048576 bytes) [37 bytes]
DEBUG    websockets.server:protocol.py:169 = connection is CLOSING
DEBUG    websockets.server:protocol.py:731 > EOF
DEBUG    websockets.server:connection.py:890 x half-closing TCP connection
```

but now we're able to negotiate the full response and subsequently close gracefully.

### DIFF
--- a/tests/v2/test_v2_session_lifecycle.py
+++ b/tests/v2/test_v2_session_lifecycle.py
@@ -1,17 +1,26 @@
 import asyncio
+import logging
 from typing import AsyncIterator, Awaitable, Callable, TypeAlias, TypedDict
 
+import msgpack
+import nanoid
 import pytest
-from websockets import ConnectionClosedOK
+from websockets import ConnectionClosed, ConnectionClosedOK
 from websockets.asyncio.server import ServerConnection, serve
 from websockets.typing import Data
 
 from replit_river.common_session import SessionState
 from replit_river.messages import parse_transport_msg
 from replit_river.rate_limiter import RateLimiter
-from replit_river.rpc import TransportMessage
+from replit_river.rpc import (
+    ControlMessageHandshakeRequest,
+    ControlMessageHandshakeResponse,
+    HandShakeStatus,
+    TransportMessage,
+)
 from replit_river.transport_options import TransportOptions, UriAndMetadata
-from replit_river.v2.session import Session
+from replit_river.v2.client import Client
+from replit_river.v2.session import STREAM_CLOSED_BIT, Session
 
 
 class _PermissiveRateLimiter(RateLimiter):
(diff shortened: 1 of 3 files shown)


tests/v2/test_v2_session_lifecycle.py
```diff
@@ -65,7 +76,13 @@ async def handle(websocket: ServerConnection) -> None:
                     state["ipv4_laddr"] = pair
         serve_forever = asyncio.create_task(server.serve_forever())
         yield None
-        serve_forever.cancel()
+        server.close()
+        await server.wait_closed()
+        try:
+            await serve_forever
+        except asyncio.CancelledError:
+            # We expect server.close() to emit CancelledError
```
jackyzha0, line 84: this block seems to imply serve_forever emits CancelledError, not server.close

jackyzha0 (commented):
it seems to me that we should just kill the session in this case? its weird that we break the connection on max size exceeded but allow it on reconnect

jackyzha0 (approved):
oops misread, thought .connect(max_size=None) only applied to first message

tests/v2/test_v2_session_lifecycle.py
blast-hardcheese: `serve_forever` is an `asyncio.Task`, not a method, so awaiting the task effectively joins the thread, raising any exceptions hidden within.

Since we really wanted to communicate that `serve_forever` wasn't an orphaned fork, and that should be handled already by `await wait_closed()`, I replaced it with an `assert ...done()`

PR #157, opened 2025-05-07
[bug] Adding explicit CancelledError handlers during async waiting loops

blast-hardcheese:
Why
===

Unhandled exceptions from v2 session is causing issues when forked async tasks get cancelled.

What changed
============

handle `CancelledError` for each of the variants

Test plan
=========

CI. Representative tests.


tests/v2/fixtures/raw_ws_server.py
masad-frost, line 1: We got test patterns yay

jackyzha0 (approved):

PR #163, opened 2025-08-04
Fix codegen for fields with dashes

PotentialStyx:
Why
===

Some new services require `-` `.` and `:` in field names

What changed
============

When rendering a `TypeName` its value is normalized, and a new `LiteralType` dataclass was added to avoid that normalization.

Known limitation: this new normalization only works on output types.

Test plan
=========

Run codegen on a schema with a field that has a `-` in its name

### DIFF
--- a/src/replit_river/codegen/client.py
+++ b/src/replit_river/codegen/client.py
@@ -1,6 +1,7 @@
 import json
 import re
 import subprocess
+from collections import defaultdict
 from pathlib import Path
 from textwrap import dedent
 from typing import (
@@ -529,6 +530,9 @@ def encode_type(
         # lambda x: ... vs lambda _: {}
         needs_binding = False
         encoder_names = set()
+        # Track effective field names to detect collisions after normalization
+        # Maps effective name -> list of original field names
+        effective_field_names: defaultdict[str, list[str]] = defaultdict(list)
         if type.properties:
             needs_binding = True
             typeddict_encoder.append("{")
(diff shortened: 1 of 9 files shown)


src/replit_river/codegen/typing.py
```diff
@@ -145,6 +162,12 @@ def work(
         raise ValueError("Incoherent state when trying to flatten unions")
 
 
+def normalize_special_chars(value: str) -> str:
+    for char in SPECIAL_CHARS:
+        value = value.replace(char, "_")
```
jackyzha0, line 167: this can cause name collisions (e.g. ".dark" and "-dark" would collide)

PotentialStyx: Is erroring an acceptable action if two names would collide or do you think using a more unique normalization (like using a char code as bri suggested) would be better?

jackyzha0: or just phoneticizing (e.g. _ -> underscore, . -> dot) i think erroring is fine too

tests/v1/codegen/test_rpc.py
```diff
@@ -52,6 +52,7 @@ async def test_basic_rpc(client: Client) -> None:
     res = await RpcClient(client).test_service.rpc_method(
         {
             "data": "feep",
+            "data2": datetime.now(timezone.utc),
```
masad-frost, line 55: nice

src/replit_river/codegen/client.py
```diff
@@ -653,19 +662,37 @@ def extract_props(tpe: RiverType) -> list[dict[str, RiverType]]:
...
                         })]{value}"
                     )
                 else:
+                    specialized_name = normalize_special_chars(name)
+                    effective_name = name
+                    extras = []
+                    if name != specialized_name:
+                        if base_model != "BaseModel":
+                            # TODO: alias support for TypedDict
+                            raise ValueError(
```
masad-frost, line 680: when could we hit this case?

src/replit_river/codegen/typing.py
```diff
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
 from typing import NewType, assert_never, cast
 
+SPECIAL_CHARS = [".", "-", ":", "/", "@", " ", "$", "!", "?", "=", "&", "|", "~", "`"]
```
masad-frost, line 4: what about emojis? /s

masad-frost (approved):
wew thanks

